PR #5, opened 2016-04-02
Added a simple templates project to make editing easier.

garethj-msft:
Simple non-building project added to the solution to make it easier to comprehend and edit templates.  Also couple of line endings changes that git imposes for Windows auto-line endings.


GraphODataTemplateWriter.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
```
peternied: Please do not update the Visual Studio version for this sln

LICENSE
```diff
@@ -1,25 +1,25 @@
-ODataGraphTemplateWriter
```
peternied: Looks like line endings were modified, please revert

peternied:
Could you update the readme with how the usage should be changed?

GraphODataTemplateWriter.sln
garethj-msft: per email discussion, this seems like the right change.

garethj-msft:
I updated the readme and removed the line ending-only changes per comments.

LICENSE
garethj-msft: Reverted

PR #13, opened 2016-04-22
Removing warnings about accessing 'Type' property and moving to using…

shiftylogic:
… the 'Projection' of a type for CSharp / Android / JavaScript.

The works still needs to be done for ObjC and Python, but I don't have an environment setup to test it so I'll get to it a bit later.


peternied:
RE: ObjC and Python, you just need to make sure the template output before and after this PR is identical, no need to perform any deeper testing.

shiftylogic:
Ok... git me a bit and I'll make those changes as well.

shiftylogic:
ObjC done.

Python has a crashing issue. It appears that even the master branch version is broken. I'll reach out to Travis to have a look at it.

peternied:
@shiftylogic Lets just bug the Python changes since that area is in flight

shiftylogic:
I filed an issue to track the Python Type to Projection change.

PR #20, opened 2016-05-16
Update vipr to get new annotation support.

shiftylogic:


shiftylogic:
Ran generator using old vipr and new vipr against same metadata file. Outputs are identical.

PR #24, opened 2016-06-29
For .NET: annotations + function /action overloads + excel prep + bug fixes

shiftylogic:
These changes bring support for consuming annotations provided by the metadata. The annotations will cause methods to be stripped from the the SDK classes (and thus cause a breaking change in the client). Each iteration of the client that introduces new annotations for existing entities will cause a new set of breaking changes.

Other changes include a fix to support function / action overloads from the metadata, better reserved word support for generated function parameters, a couple other tweaks in preparation for Excel, and a few bug fixes.


Templates/CSharp/Requests/MethodRequestBuilder.cs.tt
```diff
@@ -5,417 +5,119 @@
...
+namespace Microsoft.Graph
 {
     using System;
     using System.Collections.Generic;
-    using System.Text;
+    using System.IO;
 
     /// <summary>
-    /// The type <#=requestBuilderType#>.
+    /// The type DriveItemDeltaRequestBuilder.
```
ginach: Don't hard code this.

Templates/CSharp/Requests/MethodRequestBuilder.cs.tt
```diff
@@ -5,417 +5,119 @@
...
         /// </summary>
+        /// <param name="functionUrl">The request URL to </param>
         /// <param name="options">The query and header options for the request.</param>
-        /// <returns>The built request.</returns>
-        public I<#=requestType#> Request(IEnumerable<Option> options = null)
-        {
-    <#
-        if (!isAction && hasParameters)
+        /// <returns>An instance of a specific request class.</returns>
+        protected override I<#=requestType#> CreateRequest(string functionUrl, IEnumerable<Option> options)
```
ginach: Although we say we reserve the right to change stuff about the generated code without breaking I would argue changing Request() to CreateRequest() is a type of change that should be considered breaking. Every app that has integrated with the SDK would have to make updates for every request right now. Should really only be changed in a major version update of the SDK, especially since it's a naming change with no other added benefit.

ginach: Ah, nevermind. I didn't realize Request is still there too. Ignore this comment.

---
In reply to: [69044041](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/24#discussion_r69044041) [](ancestors = 69044041)

src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
```diff
@@ -42,7 +42,7 @@ private static OdcmNamespace GetOdcmNamespace(OdcmModel model)
         public static IEnumerable<OdcmClass> GetComplexTypes(this OdcmModel model)
         {
             var @namespace = GetOdcmNamespace(model);
-            return @namespace.Classes.Where(x => x is OdcmComplexClass);
+            return @namespace.Classes.Where(x => x is OdcmComplexClass && x.CanonicalName() != "microsoft.graph.Json");
```
ginach: Not relevant to your changes but why is this not microsoft.graph.json to keep consistent with our casing pattern? Should probably push back on that.

Templates/CSharp/Requests/MethodRequestBuilder.cs.tt
```diff
@@ -5,417 +5,119 @@
...
-    
-    requestConstructorParametersStringBuilder.Append(Environment.NewLine);
-    requestConstructorParametersStringBuilder.Append("                ");
-    requestConstructorParametersStringBuilder.AppendFormat("this.{0},", paramCheckedCase);
-}
+            foreach (var p in m.Parameters)
+            {
+#>
+            if (HasParameter("<#=p.Name#>"))
+                request.RequestBody.<#=p.PropertyName#> = GetParameter<<#=p.Type#>>("<#=p.Name#>");
```
ginach: Nit, wrap the if blocks in curly braces, even if only one line.

ginach: Also nit: preface functions with this. or ClassName. for static.

---
In reply to: [69045385](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/24#discussion_r69045385) [](ancestors = 69045385)

Templates/CSharp/Requests/IEntityReferenceRequest.cs.tt
```diff
@@ -17,6 +18,8 @@ namespace <#=this.GetNamespaceName(entity)#>
 
     <#=this.GetEntityReferenceRequestInterfaceDefinition(entity)#>
     {
+<# if (features.CanDelete) { #>
```
ginach: Nit: curly brace on new line. Same comment elsewhere.

ginach:
:shipit:

Templates/CSharp/Requests/MethodRequestBuilder.cs.tt
shiftylogic: Oops... that was an accident.

Templates/CSharp/Requests/MethodRequestBuilder.cs.tt
shiftylogic: Yeah... that is a protected helper function used for derived classes.

src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
shiftylogic: LOL. I agree and this comment was made. I actually probably should do a case insensitive check here.

Templates/CSharp/Requests/IEntityReferenceRequest.cs.tt
shiftylogic: I'm going to not make this change right now. I fought with myself on this for a while. The templates are hard enough to read that making that change actually makes them harder to read.

PR #25, opened 2016-06-29
corrected assembly name

chauey:


shiftylogic:
Thanks for noticing and correcting this.

PR #27, opened 2016-06-30
Fixing a bug introduced in the last set of changes plus fixing a few …

shiftylogic:
…existing warnings that were bugging me. There changes are completely benign to code generated for the SDKs


MIchaelMainer:
:+1:

PR #31, opened 2016-07-06
Changes to generate OneDrive SDK, moving logic from edmx to config

ginach:
- Add OneDrive metadata file. Consumer doesn't (yet) have a static metadata so we need to keep generating from a file with some stuff removed but it doesn't use any LongDescription changes anymore.
- Add the Microsoft.Graph using when the file is using a different namespace.
- Allow overriding the base telemetry value (graph-dotnet-version vs onedrive-dotnet-version) via config
- Add flag to pass function parameters in the query string (delta?token=foo) instead of in parens (delta(token='foo')). Settable via config.
- oneDrive.delta has some sibling properties to the item collection that should be first-class properties. Since we don't have a way to describe this in the metadata right now (annotation in the future?) adding this via config. The format used is because I could see injecting values on non-method classes or as inputs to methods or navigation properties as well. Open to feedback on the config structure.
- oneDrive.copy is an async method that returns a monitor object and response. Add a way to specify this in config, minor updates to the logic already in the templates.

See my [SDK PR](https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/26) to see how the code generates with these changes for Graph as well as the SDK changes required for the async and parameter changes.


src/GraphODataTemplateWriter/.config/TemplateWriterSettings.json
```diff
@@ -2,28 +2,30 @@
     "AvailableLanguages": [ "Android", "ObjC", "CSharp", "Python", "JavaScript" ],
     "TargetLanguage": "CSharp",
     "Plugins": [ ],
-    "PrimaryNamespaceName": "com",
+    "PrimaryNamespaceName": "oneDrive",
```
ginach: Doing this PR into a OneDrive branch. For the Graph branches these config changes would be reverted, except for getting rid of the Shared mapping below.

Templates/CSharp/Model/ComplexType.cs.tt
```diff
@@ -26,14 +26,19 @@ if (!complex.IsAbstract && complex.Derived != null && (complex.Base == null || c
...
 
-namespace <#=complex.Namespace.GetNamespaceName()#>
+namespace <#=namespaceName#>
 {
     using System;
     using System.Collections.Generic;
     using System.IO;
     using System.Runtime.Serialization;
+<# if (!this.IsGraphNamespace(namespaceName)) { #>
+    using Microsoft.Graph;
```
shiftylogic: Rather than doing something like this, would it not work to just fully qualify types where this matters?

src/GraphODataTemplateWriter/.config/TemplateWriterSettings.json
```diff
@@ -2,28 +2,30 @@
...
     "TemplatesDirectory": "../../../../Templates",
     "DefaultFileCasing": "UpperCamel",
     "CustomFlags": [ "python2" ],
     "LicenseHeader": [
         "Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information."
     ],
+    "AdditionalMethodProperties": {
+        "oneDrive.delta": {
+            "CollectionResponse": {
+                "@delta.token": "Edm.String",
```
shiftylogic: This feels like a hack for just onedrive service functionality. Is there any other services on the Graph that need this? Shouldn't we focus on getting the correct changes to OData and / or annotations to support this instead of putting a hack in place?

shiftylogic:
Added a couple comments.

:clock3:

Templates/CSharp/Model/ComplexType.cs.tt
ginach: That would work too, though it would be a much bigger change as we'd have to update every entity, base class, collectionpage, etc reference in each of the templates.

src/GraphODataTemplateWriter/.config/TemplateWriterSettings.json
ginach: Sure, there could be plenty. This is a delta concept in general, which would apply to the other usages of delta in Graph. There are also some navigation property requests that have required parameters that aren't discoverable at all in metadata. We could follow this pattern for those.

Agreed this is a hack and we should do this using something like annotations. And it would be removed when there's a better answer. Not sure it's on any radar and I don't want to wait 6 months for that to happen before we consolidate all this logic. Fixing the duplication issues and reducing maintenance costs is so valuable. @rgregg, thoughts on feasibility of fixing this via annotations soon?

ginach: Worth calling out this isn't actually a new hack, just moving it into the config (it's much better in config). The old approach was to have a LongDescription value set to "MethodResponse" on a manually-created entity in the metadata. Then the delta method was updated to return that method response object and convert it to the actual return values for delta.

Templates/CSharp/Model/ComplexType.cs.tt
ginach: So, yes but phase 2 change?

Templates/CSharp/Requests/IMethodAsyncMonitor.cs.cs
```diff
@@ -0,0 +1 @@
+ErrorGeneratingOutput
```
iambmelt: What's this?

cdmayer: Looking into this.

cdmayer:
I don't think I can commit to Gina's branch. Maybe we can go ahead and merge this, then I will fix the issues in another PR? My team is the only one consuming the onedrive branch so I think that would be fine.

PR #36, opened 2016-07-09
Python

travis032654:


Templates/Android/extensions/EntityWithReferencesRequest.java.tt
```diff
@@ -0,0 +1,21 @@
+<# // Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information. #>
```
iambmelt: Let's make sure we figure out what's going on here [or roll it back] before we `master` it

PR #47, opened 2016-07-17
Remove duplicate capabilities from the list to prevent crashes.

dianambb:
For some entities, the Capabilities element has duplicates of the same annotation. This causes a crash in Vipr. Remove the duplication annotations before iterating though them.


shiftylogic:
This is not the right place to fix this. If the issue is in the projections, we should file a bug against VIPR.

shiftylogic:
If you have a clean repo of this, can you file an issue [here](https://github.com/Microsoft/Vipr/issues). I will reach out the team that maintains VIPR and make sure they see this.

PR #49, opened 2016-07-20
Redoing Pr/31 - Changes to generate OneDrive SDK, moving logic from edmx to config

cdmayer:
Basically a copy of #31 (thanks @ginach!). Some changes were needed to the PR but with no access, we need to do this end-around.


cdmayer:
@shiftylogic This should be very familiar. Fixed the MethodAsyncMonitor stuff. Please take a look.

shiftylogic:
@cdmayer The original PR was being merged to a onedrive branch. Do we want these in master?

cdmayer:
...no. Thank you for pointing that out. Give me a bit to work that out and I'll poke again.

PR #62, opened 2016-09-26
Typescript definitions

dan-silver:
Generates https://github.com/microsoftgraph/msgraph-typescript-typings/blob/master/microsoft-graph.d.ts


src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeHelperTypeScript.cs
```diff
@@ -0,0 +1,72 @@
...
+            foreach (var member in _enum.Members)
+            {
+                if (enumVals == null)
+                {
+                    enumVals = '"' + member.Name + '"';
+                } else
+                {
+                    enumVals += " | " + '"' + member.Name + '"';
+                }
+            }
```
shiftylogic: This entire function can be replaced with:
  return _enum.Members.Select(s => "\"" + s + "\"").Join(" | ");

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeHelperTypeScript.cs
```diff
@@ -0,0 +1,72 @@
...
+                    break;
+                case "Boolean":
+                    typeStr = "boolean";
+                    break;
+
+            }
+            return (prop.IsCollection) ? "[" + typeStr + "]" : typeStr;
+            
+        }
+        public static String UpperCaseFirstChar(String s)
```
shiftylogic: Why is this exact function in 2 different classes?

src/GraphODataTemplateWriter/PathWriters/TypeScriptPathWriter.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+
+            @namespace = ConfigurationService.Settings.NamespaceOverride;
+            return folderName != null ? string.Format("{0}.{1}", @namespace, folderName)
+                                         : @namespace;
+        }
+
+        private string CreatePathFromNamespace(string @namespace)
+        {
+            var splittedPaths = @namespace.Split('.');
```
shiftylogic: "splitPaths"?

dan-silver: This code was copy-pasted from another path writer. It looks like they all have this duplicate code unfortunately. I'd prefer for the TS file to not diverge from the others so when it's cleaned up we can easily find duplicates.

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeHelperTypeScript.cs
shiftylogic: Sorry. Aggregate not join, and the syntax is slightly different. I was crossing my .NET with some JavaScript.

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeHelperTypeScript.cs
dan-silver: One was referenced in t4 through the code writer and the other was a private helper in the TypeHelper. I've refactored & merged them.

src/GraphODataTemplateWriter/PathWriters/TypeScriptPathWriter.cs
dan-silver: Just following the pattern of the other templates :) I've updated the name

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeHelperTypeScript.cs
dan-silver: Thanks! Got it working with return _enum.Members.Select(m => "\"" + m.Name + "\"").Aggregate((cur, next) =>  cur + " | " + next);

PR #65, opened 2016-10-10
Update base class of Post/Get method request builders to Action/Funct…

mikemcl-msft:
Update base class of Post/Get method request builders to Action/Function request builders.


shiftylogic:
This change already came in with the big Android 1.1 update.

PR #66, opened 2016-10-10
Add support for more Android specific type matching.

mikemcl-msft:
Add Duration, TimeOfDay, Json, and Stream type matching for Android.

### DIFF
--- a/src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs
+++ b/src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs
@@ -37,6 +37,13 @@ namespace Microsoft.Graph.ODataTemplateWriter.CodeHelpers.Android
                     return "com.microsoft.graph.model.DateOnly";
                 case "Binary":
                     return "byte[]";
+                case "Stream":
+                    return "java.io.InputStream";
+                case "Duration":
+                    return "javax.xml.datatype.Duration";
+                case "TimeOfDay":
+                case "Json":
+                    return "String";
                 default:
                     return @type.Name.ToUpperFirstChar();
             }


src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs
```diff
@@ -37,6 +37,13 @@ public static string GetTypeString(this OdcmType @type)
                     return "com.microsoft.graph.model.DateOnly";
                 case "Binary":
                     return "byte[]";
+                case "Stream":
+                    return "java.io.InputStream";
+                case "Duration":
+                    return "javax.xml.datatype.Duration";
+                case "TimeOfDay":
+                case "Json":
```
shiftylogic: The Json type part of this change is handled by the big Android 1.1 update.

src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs
```diff
@@ -37,6 +37,13 @@ public static string GetTypeString(this OdcmType @type)
                     return "com.microsoft.graph.model.DateOnly";
                 case "Binary":
                     return "byte[]";
+                case "Stream":
```
shiftylogic: @iambmelt 
Android ignores streams at the property level and pushes developers to call a "getContent" method on the parent request builder. We aren't surfacing any property of type stream on the models.

src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs
```diff
@@ -37,6 +37,13 @@ public static string GetTypeString(this OdcmType @type)
...
+                    return "java.io.InputStream";
+                case "Duration":
+                    return "javax.xml.datatype.Duration";
```
shiftylogic: @iambmelt 
Have you validated that this type succeeds the de-serialization? If this works, can you please extract this change and the TimeOfDay change and re-submit with appropriate unit tests?

src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs
iambmelt: This is correct - this relevant exclusion can be found in [`BaseModel.template.tt#CreatePropertyDef`](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/9c434fc35e09fecc2355469d7fd38edd6aef69db/Templates/Android/BaseModel.template.tt#L637). The deferred implementation is [here](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/9c434fc35e09fecc2355469d7fd38edd6aef69db/Templates/Android/generated/BaseEntityRequestBuilder.java.tt#L96-L98), with its interface found [here](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/9c434fc35e09fecc2355469d7fd38edd6aef69db/Templates/Android/generated/IBaseEntityRequestBuilder.java.tt#L70).

src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs
iambmelt: This is true, although [my implementation](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/master/src/GraphODataTemplateWriter/CodeHelpers/Android/TypeHelperAndroid.cs#L38-L39) differs and relies upon the [gson dependency](https://github.com/microsoftgraph/msgraph-sdk-android/blob/master/graphsdk/build.gradle#L62). @mikemcl-msft With justification, I could be convinced that there may be a reason not lock this implementation to this dependency and would happily accept/make such a change

cbales:
Sorry for the delay, @mikemcl-msft 
These changes have been added since your PR. Thanks for your contribution!

PR #85, opened 2017-04-19
Many changes, needs review

MIchaelMainer:
[53036e5](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/commit/53036e5bd49ec52a98821ec1be8ab0c99a2b7d38) - Turned on logging for template src, set default to config to .Net, and made change a change in OdcmModelExtensions, GetEntityReferenceTypes around filtering out reference collections. This last one has the great opportunity to impact generation and needs another set of eyes in case we see downstream impacts at generation.

[c7a70ec](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/commit/c7a70eca601af2577eb8cdff6948e58816e35b7d) - Add overload to GetSanitizedPropertyName in case we generate a property name that collides with the name of the type.

6be8c55 - JsonProperty attribute, changed the Required property to FQN form to avoid collisions with property named Required.

cdd7ef9 - Fixed issue where Edm.Boolean was generated into Bool instead of bool.

[9850b83](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/commit/9850b8342e0d7649ab4a6ae69ce197dc2f0dd2a6) - Fixed scenario where we generated collisions when there were entities whose name ended with "Request". The collision occurred for EventMessage and **EventMessageRequest** entities. A request object is generated for both of them, as well as matching models. This led to the generation of request objects called **EventMessageRequest** and EventMessageRequestRequest. The bold entity and request object names collided with each other.

[88582s1](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/commit/88582d1bbbfbab4e6feb94e865246c52602679ec) - we now support the generation of enum flags.

### DIFF
--- a/Templates/CSharp/Base/EntityRequest.Base.template.tt
+++ b/Templates/CSharp/Base/EntityRequest.Base.template.tt
@@ -197,8 +197,16 @@ public string GetEntityCreateAsyncMethod(OdcmClass odcmClass)
 
     var templateWriterHost = (CustomT4Host)Host;
     var templateWriter = (CodeWriterCSharp)templateWriterHost.CodeWriter;
+
     var entityName = this.GetEntityNameString(odcmClass);
-    var lowerCaseEntityName = entityName.ToLowerFirstChar();
+    
+	// Special case for when an entity name ends with "Request". Associated with the change in EntityType.cs.tt
+	if (entityName.EndsWith("Request"))
+	{
+		entityName = String.Concat(entityName, "Object");
+	}
+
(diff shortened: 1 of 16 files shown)


MIchaelMainer:
I'm going to

1. make the suggested fix to EntityRequest.cs.tt
2. run the tests on the built .Net code files before merging this PR.

MIchaelMainer:
Passes test when built for the last metadata used to generate the client. 11/2016. We are good to start generating against latest.

PR #87, opened 2017-04-28
Android deltalink

cbales:


Templates/Android/generated/BaseEntityCollectionRequestBuilder.java.tt
```diff
@@ -52,6 +52,15 @@ if (currentTypeProjection != null) {
         return new <#=TypeCollectionRequestBuilder(method)#>(getRequestUrlWithAdditionalSegment("<#=MethodFullName(method)#>"), getClient(), null<#=MethodParametersValues(method)#>);
     }
 <#
+	// Add a method to support delta query by string (opaque deltaLink)
```
MIchaelMainer: I think this is fine. I understand this is used to generate a request builder for OData functions and actions bound to collections. Instead of the method name, we are appending the deltalink to the URL.  I might understand this.

Templates/Android/generated/BaseMethodCollectionPage.java.tt
```diff
@@ -5,12 +5,26 @@
 <#host.TemplateName = BaseTypeCollectionPage(c);#>
 <#=writer.WriteHeader()#>
 <#=CreatePackageDef(host)#>
+<# 
+	var deltaPage = false;
+	if (TypeName(c).Length > 6 && TypeName(c).Substring(TypeName(c).Length-5) == "Delta") {
```
MIchaelMainer: Per conversation, and as far as I recall, you mentioned that this is used to determine whether the collection page is a delta page as there are different types of delta pages. This looks like it is using some convention where "Delta" is suffix of name. Could you add comments describing what this is in case we need to revisit this?

Templates/Android/generated/IBaseMethodCollectionPage.java.tt
```diff
@@ -5,8 +5,22 @@
 <#host.TemplateName = IBaseTypeCollectionPage(c);#>
 <#=writer.WriteHeader()#>
 <#=CreatePackageDef(host)#>
+<# 
+	var deltaPage = false;
+	if (TypeName(c).Length > 6 && TypeName(c).Substring(TypeName(c).Length-5) == "Delta") {
```
MIchaelMainer: Same as my other comment.

Templates/Android/generated/BaseMethodCollectionPage.java.tt
MIchaelMainer: Is there any chance that this could be true for something other than a delta page? Could something be named to make this true and not be a delta page?

PR #91, opened 2017-07-10
Bugfix for serializing enums in PHP

cbales:
We have to do this manually because Enum is a custom class not handled by PHP Core


Templates/Android/generated/BaseEntityCollectionReferenceRequest.java.tt
```diff
@@ -27,15 +27,15 @@
         final String requestUrl = getBaseRequest().getRequestUrl().toString();
         final ReferenceRequestBody body = new ReferenceRequestBody(getBaseRequest().getClient().getServiceRoot() + "/<#=prop#>/" + new<#=TypeName(c)#>.id);
         new <#=TypeWithReferencesRequestBuilder(c)#>(requestUrl, getBaseRequest().getClient(), /* Options */ null)
-            .buildRequest()
+            .buildRequest(getBaseRequest().getOptions())
```
dan-silver: Is there ever a case when `getBaseRequest()` will return null that we should guard against?

cbales: This is just a URL builder. If it returned null we'd have a bug in the generated code

PR #92, opened 2017-07-11
Return the entity from setODataType

cbales:


dan-silver:
Is there any test coverage of `setODataType()` that needs to be updated as well?

cbales:
Tests aren't currently auto-generated, so this would be in the [msgraph-sdk-php](https://github.com/microsoftgraph/msgraph-sdk-php) repo. Currently, there are only tests for non-string getters and setters, except for ```getProperties();``` Once we auto-generate, it will make more sense to include this.

PR #96, opened 2017-07-28
Android custom file updates

cbales:
There are currently a list of files we manually track when we regenerate the Android SDK that have to be hand-merged. Moving this code into the generator allows us to automate this process.

I have added methods for adding import statements, methods, and overwriting code. These are currently only implemented on the templates that require them.

Each of these methods takes advantage of CodeHelper classes that define the update contents to be made to each file, in accordance with the field's name. This allows us to "automagically" iterate through and apply rules to templates without hardcoding which templates to use in the BaseModel template file.


cbales:
There are also two files (ChunkedUploadRequest and ChunkedUploadResult) that need to be generated verbatim each time the generator is run. They are not described by the metadata.

PR #109, opened 2017-11-18
Dev update method request

MIchaelMainer:
Verified through diff that these changes do not affect the existing client library. This should fix the generation issues.

### DIFF
--- a/Templates/CSharp/Requests/MethodRequest.cs.tt
+++ b/Templates/CSharp/Requests/MethodRequest.cs.tt
@@ -298,7 +298,7 @@ namespace <#=method.Namespace.GetNamespaceName()#>
         /// </summary>
         /// <param name="<#=returnEntityParameter#>">The <#=returnEntityType#> object set with the properties to update.</param>
         /// <returns>The task to await for async call.</returns>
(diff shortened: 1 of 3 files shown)


Templates/CSharp/Requests/MethodRequest.cs.tt
```diff
@@ -298,7 +298,7 @@ namespace <#=method.Namespace.GetNamespaceName()#>
         /// </summary>
         /// <param name="<#=returnEntityParameter#>">The <#=returnEntityType#> object set with the properties to update.</param>
         /// <returns>The task to await for async call.</returns>
-        public <#=methodReturnType#> PatchAsync(<#=returnEntityType#> <#=returnEntityParameter#>)
+        public <#=methodOverloadReturnType#> PatchAsync(<#=returnEntityType#> <#=returnEntityParameter#>)
```
cbales: These two variables seem to have the same value throughout the file (assignment at line 33)

MIchaelMainer: They are the same unless the return type is a collection.

MIchaelMainer:
Generated with these changes against the last version of the metadata that we used to generate the client library. No code changes were generated.

PR #111, opened 2018-01-12
Added the December 2017 metadata.

MIchaelMainer:
This metadata should be the target metadata for the Jan/Feb 2018 release of the graph client libraries.


MIchaelMainer:
@HRook This is our code file generator.

PR #115, opened 2018-01-28
fix mutable default args

j0hnsmith:
@dmahugh This is a major bug, fixes https://github.com/microsoftgraph/msgraph-sdk-python/issues/27.

All instances share state, that's not intended. See linked issue for full description.


cbales:
Closing as we no longer support Python

dmahugh:
A quick clarification, FYI - we no longer support **the Python SDK**, and our recommended approach for Python developers is to use Requests to make REST calls to Graph. Here's some info about how to get started on that approach: [Get started with Microsoft Graph in a Python app](https://developer.microsoft.com/en-us/graph/docs/concepts/python).

PR #121, opened 2018-02-14
Implicit containment

cbales:
Handles the case where a navigation property references an implicit EntitySet of a singleton.


cbales:
It will not. I made a separate branch "implicit-containment-java" that is a branch off of "Java". Once I merge the Java branch we won't have to do this anymore. Sorry for the confusion.

MIchaelMainer:
You rock!!!

PR #130, opened 2018-04-24
Added metadata for April 2018 client library updates.

MIchaelMainer:
stagingv1.0_2018_04_24_source.xml - the non-processed metadata from source
stagingv1.0_2018_04_24_noannots_noIntune.xml - Intune entities and annotations are removed.
stagingv1.0_2018_04_24_noannots_Intune.xml - Intune entities are included and annotations are removed.
stagingv1.0_2018_04_24_descriptions_noIntune.xml - Description annotations are added; removed Intune entities.
stagingv1.0_2018_04_24_descriptions_Intune.xml - Description annotation are added; keep Intune.


MIchaelMainer:
@cbales Should we use this opportunity to identify new Intune types and add them to our Intune preprocessor step?

cbales:
We only need an intune-free version with descriptions. I would remove the intune name from all the others to remove verbosity and delete the noannots_nointune version. If we are also going to start including the pre-processed metadata files, we might want to consider putting these into folders.

MIchaelMainer:
- [ ] Run the metadata through the LongDescriptionsAdder to clean out the bad characters. Will want to update this in the pipeline build definitions.

MIchaelMainer:
> We only need an intune-free version with descriptions. I would remove the intune name from all the others to remove verbosity and delete the noannots_nointune version. If we are also going to start including the pre-processed metadata files, we might want to consider putting these into folders.

Addressed this although I did keep Intune-ful version in the name by accident. No more bad chars. Updated longdescriptionadder to work in the pipeline (fixed incorrect paths).

PR #135, opened 2018-06-20
Fixing BaseMethodBody java template files to use import statements co…

deepak2016:
…rrectly

This is part of the effort to reduce memory footprint for java sdk. As per the findings, generic import statements are contributing to large memory usage. This template generates ~450 files. With this fix, these 450 files will start having optimized import statements, resulting in lesser memory footprint.


deepak2016:
Yes, it is being used by other templates which we have not fixed. This is the method which adds generic import statements in all files.

PR #137, opened 2018-06-22
Java template fix to reduce memory footprint - continued

deepak2016:
Fixing import statements for BaseMethodBodyRequest template
Fixing import statements for IBaseMethodBodyRequest template
Together these templates impact ~800 files and these files will start getting optimized import statements with this fix.
Testing:
After generating files using these templates, java sdk has started building with 3GB RAM.

### DIFF
--- a/Templates/Java/BaseJavaModel.template.tt
+++ b/Templates/Java/BaseJavaModel.template.tt
@@ -567,6 +567,58 @@
 		return sb.ToString();
 	}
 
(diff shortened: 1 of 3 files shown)


deepak2016:
Find attached zip file with files having before and after changes.







[Files-with-before-and-after-changes.zip](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/files/2127199/Files-with-before-and-after-changes.zip)

PR #138, opened 2018-06-25
Java template fix to reduce memory footprint - continued

deepak2016:
Fixing import statements for BaseMethodRequestBuilder template
Fixing import statements for BaseEntityCollectionResponse template
Together these templates impact ~600 files and these files will start getting optimized import statements with this fix.
Testing:
After generating files using these templates, java sdk has started building with 2.5GB RAM.

### DIFF
--- a/Templates/Java/BaseJavaModel.template.tt
+++ b/Templates/Java/BaseJavaModel.template.tt
@@ -556,6 +556,17 @@
 		return getPrefixForModels(propertyType);	
 	}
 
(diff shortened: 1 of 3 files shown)


deepak2016:
[Files-With-And-Without-ImportChanges.zip](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/files/2132796/Files-With-And-Without-ImportChanges.zip)

PR #143, opened 2018-08-15
Added new command line exe for SDK generation

darrelmiller:


MIchaelMainer:
Ran this against last source metadata for .Net and the results were the same. No changes to functionality.

GraphSDKGenerator.nuspec
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package >
+  <metadata>
+    <id>GraphSDKGenerator</id>
+    <version>$version$</version>
+    <authors>Microsoft</authors>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Tool for generating Graph SDKs</description>
+    <releaseNotes></releaseNotes>
+    <copyright>Copyright Microsoft</copyright>
```
MIchaelMainer: `<copyright>© Microsoft Corporation. All rights reserved.</copyright>`

Per NuGet compliance

GraphSDKGenerator.nuspec
```diff
@@ -0,0 +1,22 @@
...
+    <licenseUrl>http://msdn.microsoft.com/en-US/cc300389</licenseUrl>
+    <frameworkAssemblies>
+      <frameworkAssembly assemblyName="System" targetFramework="net45" />
+    </frameworkAssemblies>
+  </metadata>
+  <files>
+    <file src="src\GraphSDKGenerator\bin\Release\GraphSDKGenerator.exe" target="tools" />
+    <file src="src\GraphSDKGenerator\bin\Release\*.dll" target="tools" />
```
MIchaelMainer: `src\GraphSDKGenerator\bin\Release\*.dll` is going to include all of those dlls that get generated on the fly for each template. I think we'll want to be explicitly inclusive in this list.

src/GraphODataTemplateWriter/Settings/ConfigurationService.cs
```diff
@@ -11,18 +11,27 @@ public static class ConfigurationService
...
             _configurationProvider = configurationProvider;
+            if (!String.IsNullOrEmpty(targetLanguage))
+            {
+                ConfigurationService.targetLanguage = targetLanguage;
+            }
         }
 
         private static TemplateWriterSettings LoadSettingsForLanguage()
         {
             TemplateWriterSettings mainTWS = _configurationProvider.GetConfiguration<TemplateWriterSettings>();
```
MIchaelMainer: Is the default language really CSharp or is what was last checked in to the template writer settings? I propose that we make CSharp the default, and ignore template writer settings.

src/GraphSDKGenerator/Options.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        Minimal,
+        Info,
+        Debug, 
+        Trace
+    }
+
+    class Options
+    {
+        [Option('l', "language", Default = "CSharp", HelpText = "Desired output language")]
```
MIchaelMainer: `HelpText = "The target language for the generated code files. The values can be: Android, Java, ObjC, CSharp, PHP, Python, TypeScript, or GraphEndpointList"`

src/GraphSDKGenerator/Program.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        private static void GenerateSDK(Options opts)
+        {
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+
+            SetupLogging(opts.Verbosity);
+
+            var edmxContents = MetadataResolver.GetMetadata(opts.Metadata);
```
MIchaelMainer: Change edmxContents to csdlContents per:

image

src/GraphSDKGenerator/MetadataResolver.cs
```diff
@@ -0,0 +1,53 @@
...
+    using System.Net.Http;
+    using System.Text;
+
+        internal class MetadataResolver
+        {
+            internal static Logger Logger => LogManager.GetLogger("MetadataResolver");
+
+            public static string GetMetadata(string metadataPath)
+            {
+                string edmxContents = "";
```
MIchaelMainer: Change edmxContents to csdlContents

src/GraphSDKGenerator/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("GraphSDKGenerator")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("GraphSDKGenerator")]
+[assembly: AssemblyCopyright("Copyright ©  2018")]
```
MIchaelMainer: `<copyright>© Microsoft Corporation. All rights reserved.</copyright>`

src/GraphODataTemplateWriter/Settings/ConfigurationService.cs
darrelmiller: The options class has the default of CSharp so targetLanguage should never be empty.  So we should never use the value in the settings file.

GraphSDKGenerator.nuspec
darrelmiller: It will only include the generated DLLs if we have executed the release version before doing the pack.  Once we move it into a CI build, then we shouldn't have an issue.  Also, we would have to explicitly call out all the 3rd DLLs which is a bit of a pain.  I suggest we put this off for the moment and re-visit it at some point in the near future.

MIchaelMainer:
At this point, I let's pull this in so we can rock on it. I just ran this for .Net. I first used ApiDoctor on the metadata. And then used typewriter to generate. It appears that we are losing `Nullable="false"` which is changing some of our signatures. I think APIDoctor is not writing back the Nullable attribute. 

From source and after MDS
```xml
<Action Name="add" IsBound="true">
    <Parameter Name="bindparameter" Type="Collection(microsoft.graph.workbookChart)" />
    <Parameter Name="type" Type="Edm.String" Nullable="false" Unicode="false" />
    <Parameter Name="sourceData" Type="microsoft.graph.Json" />
    <Parameter Name="seriesBy" Type="Edm.String" Nullable="false" Unicode="false" />
    <ReturnType Type="microsoft.graph.workbookChart" />
</Action>
```
After processing with ApiDoctor
```xml
<Action Name="add" IsBound="true">
    <Parameter Name="bindparameter" Type="Collection(microsoft.graph.workbookChart)" />
    <Parameter Name="type" Type="Edm.String" Unicode="false" />
    <Parameter Name="sourceData" Type="microsoft.graph.Json" />
    <Parameter Name="seriesBy" Type="Edm.String" Unicode="false" />
    <ReturnType Type="microsoft.graph.workbookChart" />
</Action>
```

This affects types generated from IEntityCollectionRequestBuilder.cs.tt:

Generated with MDS
```csharp
IWorkbookChartAddRequestBuilder Add(
    string type,
    string seriesBy,
    Newtonsoft.Json.Linq.JToken sourceData = null);
```

Generated with ApiDoctor
        
```csharp
IWorkbookChartAddRequestBuilder Add(
      string type = null,
      Newtonsoft.Json.Linq.JToken sourceData = null,
      string seriesBy = null);

```
This also affect types generated with MethodRequestBuilder.cs.tt:

Generated with MDS
```csharp
        public DirectoryObjectGetByIdsRequestBuilder(
            string requestUrl,
            IBaseClient client,
            IEnumerable<string> ids,
            IEnumerable<string> types)
            : base(requestUrl, client)
        {
            this.SetParameter("ids", ids, false);
            this.SetParameter("types", types, true);
        }
```

Generated with ApiDoctor
```csharp
        public DirectoryObjectGetByIdsRequestBuilder(
            string requestUrl,
            IBaseClient client,
            IEnumerable<string> ids,
            IEnumerable<string> types)
            : base(requestUrl, client)
        {
            this.SetParameter("ids", ids, true);
            this.SetParameter("types", types, true);
        }
```

The bigger issue here is why are the arguments getting re-ordered. That will break things. It looks like APIDoctor is doing the right thing and MDS is doing the wrong thing. 

- [x] Investigate the source of the argument reordering. This is because our darn template is not maintaining the ordering and hasn't since the beginning. We first write the non-nullable properties and then the nullable ones per [IRequestBuilder.Base, L112](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/dev/Templates/CSharp/Base/IRequestBuilder.Base.template.tt#L112), ugh. The consolation prize is this get 'fixed' by making sure that we maintain nullability.
- [ ] Why are we losing nullable=false with ApiDoctor. I didn't see this when I compared outputs of typewriter vs the generator because I still used MDS.

darrelmiller:
@MIchaelMainer We need to do some investigations on ApiDoctor.  I discovered there is a parameter `-Annotations AllAnnotations` and suspect there have been some other recent changes that may affect how metadata is manipulated.  I have no issue with this branch being merged.  It shouldn't affect our current processing.

PR #144, opened 2018-08-22
Fix ClassCastException during delete API call

NakulSabharwal:
Accepting correct method parameter during API call resolves ClassCastException.


MIchaelMainer:
@vikyding Adding you since you saw this in the sample migration.

PR #154, opened 2018-09-20
Add odata type to complex types for .Net and PHP

MIchaelMainer:
@odata.type property is now serialized for all complexTypes. We need to:
- [x] verify that this won't be a problem for the workloads. It shouldn't be an issue but I'd like to ask an OData expert.

This will fix https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/294.

This will be applicable for .Net, Java, PHP, and TypeScript.

I'll link to what this looks like in a .Net PR.

### DIFF
--- a/Templates/CSharp/Model/ComplexType.cs.tt
+++ b/Templates/CSharp/Model/ComplexType.cs.tt
@@ -44,7 +44,10 @@ namespace <#=complex.Namespace.GetNamespaceName()#>
     {
 
 <#
-// We need to set the @odata.type property when this type has 
+// Set the @odata.type property when this type is not abstract. If problems are caused by this, 
+// then we should consider only setting @odata.type when this types base class is an abstract class
+// referenced as the type for a property. That is really the only time when the API will need the
+// disambiguation.
 if (!complex.IsAbstract)
 {
 #>        public <#=complexTypeName#>()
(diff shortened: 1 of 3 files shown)


MIchaelMainer:
One issue for .Net is that if we have any custom partials with a constructor defined for it, we can't the change produced by this template updated.
For example, here is a custom partial for [PlannerChecklistItems (custom)](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/dev/src/Microsoft.Graph/Models/Extensions/PlannerChecklistItems.cs). It has a constructor defined in it. Here is the other partial for [PlannerChecklistItems (generated)](https://github.com/microsoftgraph/msgraph-sdk-dotnet/blob/dev/src/Microsoft.Graph/Models/Generated/PlannerChecklistItems.cs). We will have a collision if we create a constructor in the generated constructor like this:

        public PlannerChecklistItems()
        {
            this.ODataType = "microsoft.graph.plannerChecklistItems";
        }

darrelmiller:
@MIchaelMainer All the more reason to work to towards phasing out customizations of the models via partials.  I think extensions methods are really the only safe mechanism and even they can have naming clashes.

MIchaelMainer:
I've updated this to only add the constructor when we need to initialize the odatatype.

You can now review the results of this in https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/302.

PR #160, opened 2018-10-18
Choco pack and versioning script made more robust

MIchaelMainer:
The existing script in the build task didn't do its job. This script reads the latest published package version and updates the new package version based on assembly version and/or the last published version.


MIchaelMainer:
This script will eventually be moved out of build and ran in a release pipeline.

PR #166, opened 2018-11-08
Add generation options

MIchaelMainer:
This allows us to use typewriter in different ways in our build pipeline. There are now three generation options:
* Clean, annotate, and generate output metadata from an input metadata file.
* Generate code files from arbitrary input metadata without cleaning or annotating the metadata.
* Clean and annotate an input metadata before generating code files.


MIchaelMainer:
Ping @pwombwa @darrelmiller @deepak2016 @muthurathinam @jobala for code review. I need this in a build task.

PR #168, opened 2018-12-04
Use both description and long description annotations

MIchaelMainer:

MarkdownScanner to ApiDoctor appeared to change the way doc comments are parsed into the object model. This change addresses that change in the templates.

This will address the issue of the missing descriptions in https://github.com/microsoftgraph/msgraph-sdk-php/pull/116

### DIFF
--- a/Templates/PHP/Model/EntityType.php.tt
+++ b/Templates/PHP/Model/EntityType.php.tt
@@ -69,9 +69,9 @@ foreach(var property in entity.Properties.Where(prop => prop.Type.GetTypeString(
 
      /** 
      * Gets the <#=propertyName#>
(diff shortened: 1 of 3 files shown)


Templates/PHP/Model/EntityType.php.tt
```diff
@@ -140,9 +140,9 @@ if (property.Type.GetTypeString()[0] == '\\') { #>
     
     /**
     * Sets the <#=propertyName#>
-<# if (property.LongDescription != null) {
+<# if (property.LongDescription != null || property.Description != null) {
```
pwombwa: Duplication of null Description/LongDescription check since it's being handled by GetSanitizedLongDescription

PR #169, opened 2018-12-04
Use description and long description annotations for Java

MIchaelMainer:

MarkdownScanner to ApiDoctor appeared to change the way doc comments are parsed into the object model. This change addresses that change in the templates. Want to make sure that we don't miss descriptions. 

I branched this from dev. One thing I didn't understand is why did I generate with the import * everything declarations instead of what I see checked into the repo. Is there a branch that needs to be pulled into dev that you've been using for generation?

image


deepak2016:
Hi Michael, as extension folders never get overwritten into the SDK, we didn't fix templates in extension folder. I had directly fixed sdk files in the repo for them.
However, as generated files always gets overwritten, so we fixed templates for generated files.

PR #170, opened 2018-12-05
Use description and long description annotations for TS

MIchaelMainer:

MarkdownScanner to ApiDoctor appeared to change the way doc comments are parsed into the object model. This change addresses that change in the templates.

This will add many more jsdoc comments.

### DIFF
--- a/Templates/TypeScript/src/entity_types.ts.tt
+++ b/Templates/TypeScript/src/entity_types.ts.tt
@@ -30,8 +30,8 @@ export type <#= enumType.Name.UpperCaseFirstChar() #> = <#= enumType.GetEnumValu
 export interface <#= entityType.Name.UpperCaseFirstChar() #><# if (entityType.Base != null) { #> extends <#=  entityType.Base.Name.UpperCaseFirstChar() #><# }#> {
 
 <# foreach(var prop in entityType.Properties.ToList()) { #>
-	<# if (prop.LongDescription != null) { #>
-    /** <#=prop.LongDescription#> */
+	<# if (prop.LongDescription != null || prop.Description != null) { #>
+    /** <#=prop.GetSanitizedLongDescription()#> */
 	<# } #>
 	<#=  prop.Name #>?: <#= prop.GetTypeString() #>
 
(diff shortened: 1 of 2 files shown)


src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeHelperTypeScript.cs
```diff
@@ -59,7 +59,15 @@ public static String UpperCaseFirstChar(this String s)
...
         }
-        
-        
+
+        public static string GetSanitizedLongDescription(this OdcmProperty property)
+        {
+            var description = property.LongDescription ?? property.Description;
+            if (description != null)
+            {
+                return description.Replace("<", "&lt;").Replace(">", "&gt;").Replace("&", "&amp;");
```
muthurathinam: I don't understand the need for the replacements in this. why do we need that.?

Templates/TypeScript/src/entity_types.ts.tt
```diff
@@ -30,8 +30,8 @@ export type <#= enumType.Name.UpperCaseFirstChar() #> = <#= enumType.GetEnumValu
 export interface <#= entityType.Name.UpperCaseFirstChar() #><# if (entityType.Base != null) { #> extends <#=  entityType.Base.Name.UpperCaseFirstChar() #><# }#> {
 
 <# foreach(var prop in entityType.Properties.ToList()) { #>
-	<# if (prop.LongDescription != null) { #>
-    /** <#=prop.LongDescription#> */
+	<# if (prop.LongDescription != null || prop.Description != null) { #>
```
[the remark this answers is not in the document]
MIchaelMainer: They are the same checks, but the reason is a bit different. The `<# #>`is a control block to check whether we will render the contents that follow it, while the `<#= #>` is an expression control block that we use to create the output code file content. The nice thing about having both of these is that we always get the placeholder output `/**  */`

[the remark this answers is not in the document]
MIchaelMainer: They are the same checks, but the reason is a bit different. The `<# #>`is a control block to check whether we will render the contents that follow it, while the `<#= #>` is an expression control block that we use to create the output code file content. The nice thing about having both of these is that we always get the placeholder output `/**  */`

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeHelperTypeScript.cs
MIchaelMainer: @muthurathinam We've found these characters coming from the documentation in the past. These are not valid characters in  doc comments for .Net IntelliSense files IIRC. I couldn't find a list of invalid characters for JSDoc. If this isn't a concern for the typings, I can remove. Just let me know if there is any reason to keep this for JS and if so, which characters should we remove. I'm going to work under the assumption that they aren't valid and will checkin a change.

muthurathinam: @MIchaelMainer You are right, we should escape these characters. Having these (<, >, &) characters does not effects the editor intellisense but these characters are trimmed when we generate the html pages from these jsDoc comments and when we do this escaping the comments looks as we expected

code


jsdoc_gen

PR #171, opened 2018-12-07
Merge dev into master

MIchaelMainer:
We haven't merged in a while which indicates that we may want to consider getting rid of either master or dev.


darrelmiller:
We should discuss in triage tomorrow why we have dev and master.  It's not clear to me.

PR #175, opened 2019-01-21
Vidadhee/objc model gen

dadheech115:
This pull request address the following things:

1. Remove Request builder templates for ObjC
2. Changes in model templates to generate error free Models compatible with Mac and iOS

Some examples showing difference in some of the files due to these changes :

[objc_gen.zip](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/files/2795090/objc_gen.zip)


src/GraphODataTemplateWriter/CodeHelpers/ObjC/TypeHelperObjC.cs
```diff
@@ -121,9 +125,16 @@ public static string GetTypeString(this OdcmType type)
                 case "Stream":
                     return "NSStream";
                 case "Duration":
-                    return "Duration";
+                    return "MSDate";
```
MIchaelMainer: Are you referring to this [MSDate](https://github.com/microsoftgraph/msgraph-sdk-ios/blob/dev/MSGraphSDK/MSGraphCoreSDK/Core/MSDate.m)? I don't think this is correct. This is a [ISO 8601]([url](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/cs01/odata-csdl-xml-v4.01-cs01.html#_Toc505863963)) duration, not a date. Either we provide a Duration object, or we handle converting to NSTimeInterval.

Templates/ObjC/Base/SharedObjC.template.tt
```diff
@@ -203,15 +208,15 @@ private string GetHydratedIVarFromDictionary(OdcmProperty prop)
...
+            [{0}Result addObject:{2}];
         }}
     }}
 
     _{0} = {0}Result;
         ",
         GetObjCProperty(prop),
         prop.Name,
-        prop.Projection.Type.Name.ToLowerFirstChar(),
+        "temp"+prop.Projection.Type.Name.ToUpperFirstChar(),
```
MIchaelMainer: What does the "temp" prefix do for us?

dadheech115: This code generated the getter methods for properties which are Arrays. 
A "for-each" loop iterates through the array to return all the items in these getters. 
In some of the models the individual item in array and the array itself were getting the same names. So to remove this conflict I have added a "temp" prefix. This is used inside our models to return the actual array with proper name and these temp variable do not get exposed.

MIchaelMainer: this is a nitpick, but perhaps we should use '_' per https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html

dadheech115: Variables prefixed with '_' denote the instance variables. In our case the variable is local to scope of if-block. 
This piece of code -
 `if([licenseDetails isKindOfClass:[NSArray class]]){
        for (id licenseDetails in licenseDetails){
            [licenseDetailsResult addObject:licenseDetails];
        }
    }` 

gets changed to this 
` if ([licenseDetails isKindOfClass:[NSArray class]]){
        for (id tempLicenseDetails in licenseDetails){
            [licenseDetailsResult addObject:tempLicenseDetails];
        }
    }`

If 'temp' prefix is not suitable, I can use 'single' or 'individual' also as prefix

src/GraphODataTemplateWriter/CodeHelpers/ObjC/TypeHelperObjC.cs
dadheech115: I have made change related to duration.
Have added a MSDuration class in set of hand written models and have added the code in templates to use methods from this class. 
Please have a look.

PR #176, opened 2019-01-29
Sanitize Parameter variables to fix errors

JaapSingh:
When generating the SDK, certain parameter names were not being sanitized, which resulted in errors such as myfunc(string string), as opposed to myfunc(string @string). To fix this, I sanitized the parameter .


darrelmiller:
Hi @JaapSingh Thanks for your contribution.  It is much appreciated.  Your change seems to make sense to me.  However, it would be great if you could add a short description of the problem you ran into and how using the sanitize solves the problem for you.  Thanks!

darrelmiller:
@MIchaelMainer If you are good with this fix, feel free to merge.

PR #190, opened 2019-03-21
Handle CRLF in doc annotations as code comments.

### LINKED ISSUE #187: &#xA;&#xA; in an annotation breaks the generated output
Repro schema: `test_intune_onedf` on Graph PPE.

XML snippet in question:
```xml
<Annotations Target="microsoft.graph.windows10EndpointProtectionConfiguration/rpcUnauthenticatedClientOptions">
  <Annotation Term="Org.OData.Core.V1.Description" String="This policy setting controls how the RPC server runtime handles unauthenticated RPC clients connecting to RPC servers.&#xA;&#xA;This policy setting impacts all RPC applications. In a domain environment this policy setting should be used with caution as it can impact a wide range of functionality including group policy processing itself.  Reverting a change to this policy setting can require manual intervention on each affected machine.  This policy setting should never be applied to a domain controller." />
</Annotations>
```

Generated C#:
```csharp
/// <summary>
/// Gets or sets rpc unauthenticated client options.
/// This policy setting controls how the RPC server runtime handles unauthenticated RPC clients connecting to RPC servers.

This policy setting impacts all RPC applications. In a domain environment this policy setting should be used with caution as it can impact a wide range of functionality including group policy processing itself.  Reverting a change to this policy setting can require manual intervention on each affected machine.  This policy setting should never be applied to a domain controller.
/// </summary>
```

MIchaelMainer:
Fixes #187 

HTML encoded CRLF in doc annotations gets converted to escaped chars \r\n. We need to make sure we emit the new line doc comments with ///. Updated C#, PHP, and Typescript to support this.

### DIFF
--- a/GraphODataTemplateWriter.Test/TypeHelperTests.cs
+++ b/GraphODataTemplateWriter.Test/TypeHelperTests.cs
@@ -1,0 +1,95 @@
+﻿using Microsoft.Graph.ODataTemplateWriter.CodeHelpers.CSharp;
+using Microsoft.Graph.ODataTemplateWriter.CodeHelpers.TypeScript;
+using Microsoft.Graph.ODataTemplateWriter.CodeHelpers.PHP;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Vipr.Core.CodeModel;
+
+
+namespace GraphODataTemplateWriter.Test
+{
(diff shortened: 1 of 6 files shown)


PxlBuzzard:
This looks good to me.

PR #202, opened 2019-05-28
Updating EntityRequest.Base.template with code fix for response headers issue

irvinesunday:
This PR does the following:

Updates the *EntityRequest.Base.template* with the code fix that checks response headers and status codes in the **AdditionalData** property tag of entities and throws an error (if response headers or status code properties are found). 

This fix is for task: microsoftgraph/msgraph-sdk-dotnet#426


MIchaelMainer:
- [ ] Validate in client by building and writing test.

Another thing you'll need to do is to build the client library with the updated files. Do this by copying the generated files into their appropriate directory in the client repo. Can you verify that this builds?
image

You'll also be able to validate your test at this point. If all is good, go ahead and make a PR in the SDK  repo for the updated code files.

- [ ] @MIchaelMainer Update the build agent with this built updated typewriter.exe

PR #203, opened 2019-05-29
Fixed Typescript template bugs and linting for publishing in definitely typed repo

muthurathinam:


Templates/TypeScript/src/entity_types.ts.tt
```diff
@@ -10,16 +10,16 @@
     var enumTypes               = model.GetEnumTypes();
     var complexTypes            = model.GetComplexTypes();
 #>
-// Type definitions for the Microsoft Graph API
+// Type definitions for the Microsoft Graph <VERSION_STRING>
```
MIchaelMainer: What is the intention of <VERSION_STRING>

muthurathinam: Actually, the version for definitely typed repo should be mentioned like [this](https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library), `<VERSION_STRING>` is a place holder string, we will replace this string with version before publishing.

PR #208, opened 2019-06-13
Fix linter issues for definitely typed

muthurathinam:
Fixed linting error in the typescript types generation:
* [max-line-length](https://palantir.github.io/tslint/rules/max-line-length)
* [no-empty-interface](https://palantir.github.io/tslint/rules/no-empty-interface)
* [interface-name](https://palantir.github.io/tslint/rules/interface-name)

### DIFF
--- a/Templates/TypeScript/src/entity_types.ts.tt
+++ b/Templates/TypeScript/src/entity_types.ts.tt
@@ -11,7 +11,7 @@
     var enumTypes               = model.GetEnumTypes();
     var complexTypes            = model.GetComplexTypes();
 
-    var maxLineLength = 180;
+    var maxLineLength = 120;
 #>
 // Type definitions for non-npm package microsoft-graph <VERSION_STRING>
 // Definitions by: Microsoft Graph Team <https://github.com/microsoftgraph>
--- a/Templates/TypeScript/src/entity_types.ts.tt
+++ b/Templates/TypeScript/src/entity_types.ts.tt
@@ -1,5 +1,6 @@
 <# // Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information. #>
 <#@ template debug="true" hostspecific="true" language="C#" #>
+<#@ include file="utils.tt" #>
 <#@ output extension="\\" #>
 <#
     CustomT4Host host           = (CustomT4Host) Host;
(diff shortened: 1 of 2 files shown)


Templates/TypeScript/src/entity_types.ts.tt
```diff
@@ -9,45 +10,159 @@
...
+    }
+
+    foreach(var entityType in entityTypes) {
+        var propCount = entityType.Properties.ToList().Count;
+        var entityTypeName = entityType.Name.UpperCaseFirstChar();
+        if(propCount == 0 && entityTypeName[0] == 'I') {
+#>
+// tslint:disable-next-line: interface-name no-empty-interface
+<#
+        } else if (entityTypeName[0] == 'I') {
```
MIchaelMainer: `else if (entityTypeName[0] == 'I')` should be `else if (entityTypeName[0] != 'I')` as we want to disable tslint when the interface does not start with 'I'. This doesn't account for types that are named with 'I'. I think we should disable this check on all interfaces since we don't prefix interface names with 'I'

Templates/TypeScript/src/entity_types.ts.tt
```diff
@@ -9,45 +10,159 @@
     var entityTypes             = model.GetEntityTypes();
     var enumTypes               = model.GetEnumTypes();
     var complexTypes            = model.GetComplexTypes();
+
+    var maxLineLength = 180;
```
MIchaelMainer: 180 is high, let's make it 120.

Templates/TypeScript/src/entity_types.ts.tt
```diff
@@ -9,45 +10,159 @@
...
 #>
-// Type definitions for the Microsoft Graph <VERSION_STRING>
+// Type definitions for non-npm package microsoft-graph <VERSION_STRING>
```
MIchaelMainer: Where is `<VERSION_STRING>` used?

Templates/TypeScript/src/entity_types.ts.tt
muthurathinam: Actually definitely typed repo is not allowing the interfaces starting with 'I'.

Templates/TypeScript/src/entity_types.ts.tt
muthurathinam: The allowed max length by definitely typed repo is 200, I just shortened it 180, in case tabs adds up some length to the line

Templates/TypeScript/src/entity_types.ts.tt
muthurathinam: Refer this comment in previous PR https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/203#discussion_r289249276

Templates/TypeScript/src/entity_types.ts.tt
MIchaelMainer: Ideally, we don't want comments over 80 chars in length since it requires more eye movement across the screen. 120 should be the max.

Templates/TypeScript/src/entity_types.ts.tt
MIchaelMainer: Got it.

Templates/TypeScript/src/entity_types.ts.tt
muthurathinam: Okay will fix this.

PR #219, opened 2019-09-12
Init OData.Type on complex and entity types.

MIchaelMainer:
**modified:   Templates/CSharp/Model/ComplexType.cs.tt**

Add public cstor and ODataType initialization for
non-abstract complextypes.

**modified:   Templates/CSharp/Model/EntityType.cs.tt**

Add public cstor and ODataType initialization for
non-abstract entitytypes.

modified:   test/Typewriter.Test/Given_a_valid_metadata_file_to_Typewriter.cs
modified:   test/Typewriter.Test/Resources/dirtyMetadata.xml

Address:
Fixes microsoftgraph/msgraph-sdk-dotnet-core#36

Examples:
image
image

### DIFF
--- a/Templates/CSharp/Model/ComplexType.cs.tt
+++ b/Templates/CSharp/Model/ComplexType.cs.tt
@@ -43,10 +43,8 @@ namespace <#=complex.Namespace.GetNamespaceName()#>
     public <#=classType#> <#=typeDeclaration#>
     {
 <#
-// Generate a constructor to initialize the @odata.type property when this type is not abstract and if this 
-// type's base is abstract and the base is referenced as the type of a structural property. We need this
-// to disambiguate the type of the descendant class being sent. 
-if (complex.IsBaseAbstractAndReferencedAsPropertyType() && !complex.IsAbstract)
+// Generate a constructor to initialize the @odata.type property when this type is not abstract.
+if (!complex.IsAbstract)
 {
 #>        /// <summary>
         /// Initializes a new instance of the <see cref="<#=complexTypeName#>"/> class.
--- a/Templates/CSharp/Model/EntityType.cs.tt
+++ b/Templates/CSharp/Model/EntityType.cs.tt
@@ -75,7 +75,19 @@ namespace <#=entity.Namespace.GetNamespaceName()#>
         }
     <#
         }
+		else
+		{
+	#>
 
+		///<summary>
+		/// The <#=entityName#> constructor
+		///</summary>
+        public <#=cstorTypeDeclaration#>()
+        {
+            this.ODataType = "<#=entity.FullName#>";
+        }
+	<#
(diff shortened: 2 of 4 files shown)


MIchaelMainer:
I want to make this change only after we've done a v1.0 release. I want to compare the changes it makes to the OM in isolation of other metadata driven changes.

MIchaelMainer:
@darrelmiller @peombwa @millicentachieng Can I get review?

Templates/CSharp/Model/ComplexType.cs.tt
```diff
@@ -43,10 +43,8 @@ namespace <#=complex.Namespace.GetNamespaceName()#>
     public <#=classType#> <#=typeDeclaration#>
     {
 <#
-// Generate a constructor to initialize the @odata.type property when this type is not abstract and if this 
-// type's base is abstract and the base is referenced as the type of a structural property. We need this
-// to disambiguate the type of the descendant class being sent. 
-if (complex.IsBaseAbstractAndReferencedAsPropertyType() && !complex.IsAbstract)
+// Generate a constructor to initialize the @odata.type property when this type is not abstract.
+if (!complex.IsAbstract)
```
ksikorsk: Should this be `if (complex.Base != null)`?

Templates/CSharp/Model/EntityType.cs.tt
```diff
@@ -75,7 +75,19 @@ namespace <#=entity.Namespace.GetNamespaceName()#>
...
+		else
+		{
+	#>
 
+		///<summary>
+		/// The <#=entityName#> constructor
+		///</summary>
+        public <#=cstorTypeDeclaration#>()
+        {
+            this.ODataType = "<#=entity.FullName#>";
```
ksikorsk: What would happen if developers put something else in here?

Should the `string ODataType { get; set; }` property have no setter?
`string ODataType { get; }`

Templates/CSharp/Model/ComplexType.cs.tt
MIchaelMainer: I don't think it has to be. I can't recall why having this is generally useful.

ksikorsk: It would work both ways, I was just wondering.

Templates/CSharp/Model/EntityType.cs.tt
MIchaelMainer: The service may not properly deserialize the request if a developer sets the Odata.type property to a value that matches the object. We do need at least the getter and serialization instructions. This still needs work.

MIchaelMainer: Nevermind the last comment. The complex type template has the getter/setter.

PR #220, opened 2019-09-15
Check whether this type is a generated model type or a PHP type

krizalys:
Fixed an issue where PHP models for complex types would not refer to built-in PHP types as expected. For example, without this, these models would be referring to an inexistent `Microsoft\Graph\Model\\DateTime` class, instead of PHP's `\DateTime` class.

See https://github.com/microsoftgraph/msgraph-sdk-php/pull/231.


krizalys:
@MIchaelMainer @muthurathinam What do you think about this?

krizalys:
@MIchaelMainer @muthurathinam What do you think about this?

PR #223, opened 2019-10-11
Add element reorder capability

### LINKED ISSUE #222: Create fix for re-ordered parameters
Due to autogeneration of previously manually generated CSDL for the mail API some actions have parameters that are in a different order than previously.  In the short term, we need to have Typewriter make this correction.

The following actions had their parameters re-ordered.

| Location | Name of Edm Element  |  Type of Edm Element |  Name of Parent Element  |   Error Message |
|------|------|-------|-------|--------|
|  (981, 4)  |        forward                       |    Action              | forward                       |    Parameter reordered                               |
|  (1488, 4) |      accept                        |      Action              | accept                        |      Parameter reordered                               |
|  (1493, 4) |      decline                       |      Action              | decline                       |      Parameter reordered                               |
|  (1498, 4) |      tentativelyAccept             |            Action              | tentativelyAccept             |            Parameter reordered                               |

The CSDL should look like this.... I think.
```xml
<Action Name="forward" IsBound="true">
<Parameter Name="bindingParameter" Type="microsoft.graph.post"/>
<Parameter Name="Comment" Type="Edm.String" Unicode="false"/>
<Parameter Name="ToRecipients" Type="Collection(microsoft.graph.recipient)" Nullable="false"/>
</Action>
<Action Name="accept" IsBound="true">
<Parameter Name="bindingParameter" Type="microsoft.graph.event"/>
<Parameter Name="Comment" Type="Edm.String" Unicode="false"/>
<Parameter Name="SendResponse" Type="Edm.Boolean"/>
</Action>
<Action Name="decline" IsBound="true">
<Parameter Name="bindingParameter" Type="microsoft.graph.event"/>
<Parameter Name="Comment" Type="Edm.String" Unicode="false"/>
<Parameter Name="SendResponse" Type="Edm.Boolean"/>
</Action>
<Action Name="tentativelyAccept" IsBound="true">
<Parameter Name="bindingParameter" Type="microsoft.graph.event"/>
<Parameter Name="Comment" Type="Edm.String" Unicode="false"/>
<Parameter Name="SendResponse" Type="Edm.Boolean"/>
</Action>
```

MIchaelMainer:
Fixes #222.

Enables us to reorder elements in any global metadata element. 

- [ ] Squash and merge.
- [ ] Review metadata changes to make sure we've captured all of the reordered Actions.

### DIFF
--- a/src/Typewriter/MetadataPreprocessor.cs
+++ b/src/Typewriter/MetadataPreprocessor.cs
@@ -76,8 +76,22 @@ namespace Typewriter
             AddContainsTarget("appVulnerabilityManagedDevice");
             AddContainsTarget("appVulnerabilityMobileApp");
 
-            // TODO: Reorder elements. inspect metadata so tht we capture all changes.
-
+            ReorderElements(MetadataDefinitionType.Action,
+                            "forward",
+                            new List<string>() { "bindingParameter", "Comment", "ToRecipients" },
+                            "microsoft.graph.post");
+            ReorderElements(MetadataDefinitionType.Action,
+                            "accept",
+                            new List<string>() { "bindingParameter", "Comment", "SendResponse" },
+                            "microsoft.graph.event");
+            ReorderElements(MetadataDefinitionType.Action,
+                            "decline",
+                            new List<string>() { "bindingParameter", "Comment", "SendResponse" },
+                            "microsoft.graph.event");
+            ReorderElements(MetadataDefinitionType.Action,
+                            "tentativelyAccept",
+                            new List<string>() { "bindingParameter", "Comment", "SendResponse" },
+                            "microsoft.graph.event");
 
             return xMetadata.ToString();
         }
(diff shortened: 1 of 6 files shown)


src/Typewriter/MetadataPreprocessor.cs
```diff
@@ -181,5 +200,94 @@ internal static void AddLongDescriptionToThumbnail()
...
+                XElement targetElement = null;
+
+                // Reordering elements by element Name attributes. Useful for non Action or Function.
+                if (String.IsNullOrEmpty(bindingParameterType))
+                {
+                    targetElement = results.FirstOrDefault();
+                }
+                else // We are reordering an Action or Function and must match the bindingParameterType.
+                {
+                    targetElement = results.Where(e => e.Elements().Any(a => a.Attribute("Type").Value == bindingParameterType)).FirstOrDefault();
```
andrueastman: Here, we are checking if any of the parameters/elements has the same type as the `bindingParameterType` . 
If the bindingParameter is not really of the same type as `bindingParameterType` but one of the other params has that type we might end up with a false positive.
Since the bindingParameter is typically the first parameter, would it be better to check if the type of the **first** element type matches the `bindingParameterType`. I think that may be a better invariant.

MIchaelMainer: Good catch!!! The bindingParameter is **always** the first parameter. I will fix.

MIchaelMainer: @andrueastman Made the suggested fix.

MIchaelMainer:
@peombwa @darrelmiller @andrueastman 
Can I get review as I needed to review an unnecessary review and update an ApiDoctor dependency?

PR #231, opened 2019-12-29
Create MSGraph-SDK-Code-Generator.yml

nokafor:
I added a YAML file with the metadata for this tool

These changes allow us to automatically update the Office/Graph galleries (https://developer.microsoft.com/en-us/office/gallery) with information for this tool. This means whenever this tool is referenced in the Office / Graph Developer Portals, the user will be seeing the most up-to-date information.

### DIFF
--- a/MSGraph-SDK-Code-Generator.yml
+++ b/MSGraph-SDK-Code-Generator.yml
@@ -1,0 +1,45 @@
+page_type: sample
+products:
+  - office-excel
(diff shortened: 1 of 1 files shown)


MIchaelMainer:
@nokafor I don't think we should exposes this through the portal. I don't expect this is valuable to very many customers. Please reopen if you think otherwise.

PR #235, opened 2020-02-24
Add conditional import of Microsoft.Graph when generating C# types outside of Microsoft.Graph

rofenton:
If generating types that are not in the Microsoft.Graph namespace, we need to import the Microsoft.Graph namespace in situations where objects are used from the Microsoft.Graph namespace, such as when adding Microsoft.Graph JSON Deserializer, DerivedTypeConverter, or using Microsoft.Graph types for date & time representations.

This PR adds conditional checks to the templates to add the import statement to the generated c# files in situations where Microsoft.Graph is needed, but the type being generated is not in Microsoft.Graph

### DIFF
--- a/.nuget/nuget.config
+++ b/.nuget/nuget.config
@@ -3,4 +3,7 @@
   <solution>
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
+  <config>
+    <add key="repositoryPath" value="..\packages" />
+  </config>
 </configuration>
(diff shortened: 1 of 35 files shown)


zengin:
Thanks for the proposal, but in February we were not generating for namespaces outside Microsoft.Graph.

With the namespaces support, Microsoft.Graph type references from non-Microsoft.Graph namespaces always use the fully qualified name, so this issue doesn't exist with our generated code.

Proposed solution (adding using statements) doesn't work because there are name collisions from different namespaces.

PR #247, opened 2020-06-09
Add support for logo stream property

MIchaelMainer:
Results in the following: https://github.com/microsoftgraph/msgraph-sdk-java/pull/377

- removed unused csproj entries.
- fixed missing java stream files.


src/GraphODataTemplateWriter/.config/JavaSettings.json
```diff
@@ -119,14 +119,14 @@
...
-          { "Template": "IBaseStreamRequest", "SubProcessor": "StreamProperty",                                                               "Name": "IBase<Class><PropertyType>Request", "Include": "mainLogo;thumbnailPhoto;content" },
-          { "Template": "BaseStreamRequest", "SubProcessor": "StreamProperty",                                                                "Name": "Base<Class><PropertyType>Request", "Include": "mainLogo;thumbnailPhoto;content" },
-          { "Template": "IBaseEntityRequestBuilder", "SubProcessor": "StreamProperty",                                                        "Name": "IBase<Class><PropertyType>RequestBuilder", "Include": "mainLogo;thumbnailPhoto;content" },
-          { "Template": "BaseEntityRequestBuilder", "SubProcessor": "StreamProperty",                                                         "Name": "Base<Class><PropertyType>RequestBuilder", "Include": "mainLogo;thumbnailPhoto;content" },
+          { "Template": "IEntityRequest", "SubProcessor": "StreamProperty",                                                                   "Name": "I<Class><PropertyType>Request", "Include": "mainLogo;thumbnailPhoto;content;logo" },
```
zengin, line 122: We should probably come up with a solution to either generate this list programmatically from Stream properties, or at least some mechanism to notify us when there is a new stream property in the metadata so that we can update this list.

ddyett: hopefully the build will end up failing if it's not there.

MIchaelMainer: C# doesn't use the Include list and just assumes that all stream properties are accessed the same way. I don't know if that is correct, but it hasn't caused us problems, yet. 

The build will fail if not set for Java.

PR #251, opened 2020-06-12
Fix Java generation - support methods that return enums, disambiguate entities that end in "Request"

MIchaelMainer:
The changes in this PR affect the following files:

image

The files marked in blue are changed but aren't affected by the change. They now have an additional unused import (there are many, and represent a different set of challenges). We could add a template exclusion list to the code, but that does not feel right.

## Support methods that return enums. 
The change in  BaseMethodCollectionResponse .java.tt results in us not generating setters for instance annotations for enum return types; because enum values don't have instance annotations. This only impacts the generated CalendarAllowedCalendarSharingRolesCollectionResponse.java.

image

## Disambiguate model references when the mode name ends in "Request"
The metadata has entities named timeOff and timeOffRequest. The entity named timeOff results in us generating files named models.timeOff and requests.timeOffRequest. The entity named timeOffRequest results in us generating files named models.timeOffRequest and requests.timeOffRequestRequest. This results in a name collision. The changes in this PR result in the following generated code change which address this issue:

image

### DIFF
--- a/src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
+++ b/src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
@@ -8,6 +8,7 @@ namespace Microsoft.Graph.ODataTemplateWriter.Extensions
     using Microsoft.Graph.ODataTemplateWriter.Settings;
     using Vipr.Core.CodeModel;
     using NLog;
+    using Microsoft.Graph.ODataTemplateWriter.TemplateProcessor;
 
     public static class OdcmModelExtensions
     {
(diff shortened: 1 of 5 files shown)


src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
```diff
@@ -137,6 +138,85 @@ public static IEnumerable<OdcmMethod> GetMethods(this OdcmModel model)
...
+            // At this point this is only applicable to OdcmProperty.
+            // Challenging this assumption will require a lot more investigation.
+            if (!(host.CurrentType is OdcmProperty))
+                return false; 
+
+            // We only support "Request" dismabiguation at this point. Check whether the
+            // current type ends in "Request".
+            var requestSuffix = "Request";
+            var currentTypeName = (host.CurrentType as OdcmProperty).Type.Name;
+            int index = currentTypeName.IndexOf(requestSuffix);
```
zengin, line 167: This logic may get false positives when "Request" is followed by something else, e.g. "somethingRequestSomething" (though it may be very unlikely). I think we should use `EndsWith()` instead to futureproof.

Or:
```
var shouldDisambiguate = classes.Where(entity => entity.Kind == OdcmClassKind.Entity
                                                       && entity.Name + "Request" == currentTypeName).Any();
```

But I prefer `EndsWith()` as it entails less string operations.

MIchaelMainer:
- [ ] need to add test

PR #255, opened 2020-06-25
Fix base class for PHP complex type

MIchaelMainer:
* This will update all complex type files
* e6a1c3f updates the headers in generated files. All files will be changed.


zengin:
Special thanks for fixing these before the namespace changes. I will have a working base to compare against :)

Issue #257, opened 2020-06-29
Nullable fields not communicated in types


meme:
Fields in the Graph API can often times be `null`, an example being the result from the "Create User" endpoint:

```js
{
    "@odata.context": "https://graph.microsoft.com/v1.0/$metadata#users/$entity",
    "id": "id-value",
    "businessPhones": [],
    "displayName": "displayName-value",
    "givenName": null,
    "jobTitle": null,
    "mail": null,
    "mobilePhone": null,
    "officeLocation": null,
    "preferredLanguage": null,
    "surname": null,
    "userPrincipalName": "upn-value@tenant-value.onmicrosoft.com"
}
```
(https://docs.microsoft.com/en-us/graph/api/user-post-users?view=graph-rest-1.0&tabs=http#response-1.)

However, the respective typing for `MicrosoftGraph.User`'s `jobTitle` states that it is either `string | undefined` using the `?` TS shorthand:

```ts
    jobTitle?: string;
```
(https://github.com/microsoftgraph/msgraph-typescript-typings/blob/master/microsoft-graph.d.ts#L1235.)

The response from the API (both on the docs and in practice) shows that `jobTitle` can be nullable (as can other fields). The typings should be updated to reflect nullable fields.

PR #263, opened 2020-07-20
Add Option type for NULL fields

meme:
https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/257

### DIFF
--- a/Templates/TypeScript/src/entity_types.ts.tt
+++ b/Templates/TypeScript/src/entity_types.ts.tt
@@ -25,7 +25,7 @@
 
 export as namespace microsoftgraph;
 
-type Option<T> = T | null;
+type NullableOption<T> = T | null;
 
 <#
     foreach(var enumType in enumTypes) {
@@ -105,7 +105,7 @@ export interface <#= entityTypeName #><#
 <#
     if(prop.IsNullable) {
 #>
-    <#= prop.Name #>?: Option<<#= prop.GetTypeString() #>>;
+    <#= prop.Name #>?: NullableOption<<#= prop.GetTypeString() #>>;
 <#
     } else {
 #>
@@ -176,7 +176,7 @@ export interface <#= complexTypeName #><#
 <#
     if(prop.IsNullable) {
 #>
-    <#= prop.Name #>?: Option<<#= prop.GetTypeString() #>>;
+    <#= prop.Name #>?: NullableOption<<#= prop.GetTypeString() #>>;
 <#
     } else {
 #>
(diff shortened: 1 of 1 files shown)


meme:
How can I check if property a is `null`-able?

zengin:
`prop.IsNullable`

meme:
Tried my best to follow the styling~ let me know if any changes need to be made with my indentation.

meme:
Any movement on this?

MIchaelMainer:
This adds `type Option<T> = T | undefined | null;` which then makes these changes:

image

This appears to be a breaking change as the following code results in the following message: `Argument of type '{ id: string; displayName: string; }' is not assignable to parameter of type 'UserIdentity'.
  Type '{ id: string; displayName: string; }' is missing the following properties from type 'UserIdentity': ipAddress, userPrincipalName`

```typescript
export interface UserIdentity {
    // Unique identifier for the identity.
    id: Option<string>;
    // The identity's display name. Note that this may not always be available or up-to-date.
    displayName: Option<string>;
    // Indicates the client IP address used by user performing the activity (audit log only).
    ipAddress: Option<string>;
    // The userPrincipalName attribute of the user.
    userPrincipalName: Option<string>;
}

/* Same issue */
// export interface UserIdentity {
//     // Unique identifier for the identity.
//     id: string | null | undefined;
//     // The identity's display name. Note that this may not always be available or up-to-date.
//     displayName: string | null | undefined;
//     // Indicates the client IP address used by user performing the activity (audit log only).
//     ipAddress: string | null | undefined;
//     // The userPrincipalName attribute of the user.
//     userPrincipalName: string | null | undefined;
// }

/* Original - this shows no warnings from TS */
// export interface UserIdentity {
//     // Unique identifier for the identity.
//     id?: string;
//     // The identity's display name. Note that this may not always be available or up-to-date.
//     displayName?: string;
//     // Indicates the client IP address used by user performing the activity (audit log only).
//     ipAddress?: string;
//     // The userPrincipalName attribute of the user.
//     userPrincipalName?: string;
// }

type Option<T> = T | undefined | null;

function printId(userIdentity: UserIdentity) {
    console.log(userIdentity.id);
}

// let myObj = { id: "10", displayName: "Size 10 Object", ipAddress: null, userPrincipalName: null };
let myObj = { id: "10", displayName: "Size 10 Object" };

printId(myObj);
```

@meme Can you help me understand what I see here?

meme:
This was an oversight~ since `undefined` is embedded in the type, it does not act the same as an optional field with, e.g.: `name?: string`. I have corrected this issue in my latest commit, so your example should now work correctly.

zengin:
To cover the point I mentioned in my review, dot notation in VSCode auto-complete doesn't break with this change.

MIchaelMainer:
closes #257

PR #264, opened 2020-07-21
Namespaces support - Refactor Java Generation

zengin:
Refactoring Java generation code:
- Prefix `com.` was removed accidentally due to changes in `C#` generation (we need clearer decoupling of languages)
- Every method except one was removed from `BaseJavaModel.template.tt` and added into `TypeHelperJava.cs` as extension methods. During this process, I also simplified some of the method bodies.
  - This also gives us better highlighting and Intellisense capabilities as the main logic lives in a `.cs` file instead of `.tt`.
- Updated references into extension methods in template files
- Removed Android code as we don't support it anymore and it was polluting search and replace queries with repeated code from Java generation.
- *Regression testing:* Compared the output from dev branch for V1 generation
  - The output was an exact match.

Please note that this doesn't include multiple namespaces support yet.

[AB#5032](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5032)


src/GraphODataTemplateWriter/GraphODataTemplateWriter.csproj
```diff
@@ -71,15 +67,13 @@
     <Compile Include="CodeHelpers\Python\CodeWriterPython.cs" />
```
ddyett, line 67: any specific reason we've kept python?

zengin: Darrel said that we can keep it if we can take out anything useful when we revisit Python. I didn't get a chance to judge whether there is anything useful, so I kept it for now.

Templates/Java/BaseJavaModel.template.tt
```diff
@@ -1,6 +1,7 @@
 <# // Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information. #>
 <#@ template debug="true" hostspecific="true" language="C#" #>
 <#@ output extension="\\" #>
+<#@ import namespace="Microsoft.Graph.ODataTemplateWriter.CodeHelpers.Java" #>
```
MIchaelMainer, line 4: reviewed. just adding this note for my tracking, and so you know it has been reviewed.

src/Typewriter/Options.cs
```diff
@@ -45,7 +45,7 @@ public enum GenerationMode
 
     public class Options
     {
-        [Option('l', "language", Default = "CSharp", HelpText = "The target language for the generated code files. The values can be: Android, Java, ObjC, CSharp, PHP, Python, TypeScript, or GraphEndpointList")]
+        [Option('l', "language", Default = "CSharp", HelpText = "The target language for the generated code files. The values can be: Java, ObjC, CSharp, PHP, Python, TypeScript, or GraphEndpointList")]
```
MIchaelMainer, line 48: related to this change, we need to update the readme so we no longer note this option.

src/GraphODataTemplateWriter/GraphODataTemplateWriter.csproj
MIchaelMainer: It is still useful at this time to generate models for Python. The request builders aren't useful. We've provided them to an internal team once more on as-is basis. Let's revisit when we finalize Python SDK models.

zengin:
Thanks for your review @MIchaelMainer, I have added the required Java namespace changes on a branch based on this one. So I will merge this as is and address all your comments in the next PR, to make rebasing easier.

PR #273, opened 2020-08-19
Objective-C namespaces support

zengin:
- Namespaces are converted into unique prefixes for types to prevent name collisions as Objective-C doesn't have namespaces.
- Conversion:
  - microsoft.graph portion of the namespace is converted into MSGraph, rest of the portions in namespaces are converted into PascalCase.
    - e.g. `microsoft.graph.callRecords` turns into the prefix: `MSGraphCallRecords`.
    - The type `callRecord` in `microsoft.graph.callRecords` namespace is represented as `MSGraphCallRecordsCallRecord`.
- Regression testing:
  - V1 single namespace input generates the same output for both `dev` branch and my code.
  - Beta single namespace input generates output which differs only around the `decimal` type (`MSGraphDecimal` vs `Decimal` due to `Edm` special case). It looks like this case is not handled at the moment as V1 doesn't have this problem. I will postpone fixing that as the issue is somewhat orthogonal to namespaces support. (cc: @darrelmiller @MIchaelMainer)

[AB#5030](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/5030)


zengin:
@MIchaelMainer, interesting, the test that failed in the pipeline is passing locally.

MIchaelMainer:
The updates for VIPR is causing issues for me. I'm not able to build. That's when I kicked off the test.

zengin:
My base branch doesn't have updates to VIPR. I haven't rebased since I branched off (more than a month ago).

zengin:
@MIchaelMainer, if the changes look good to you, I can merge this one to `feature-namespaces` as it is targeted now. I will have another PR rebasing that branch and merging back to `dev`. Maybe you can try running your tests from rebased `feature-namespaces`. The initial concern of merging to `dev` no longer holds as you have made `master` the release branch.

zengin:
My bad on not including the diff: https://github.com/microsoftgraph/msgraph-sdk-objc-models/commit/99a6ec00e174ea233b7cf3b98a838743818c3e04

I skipped it because single namespace output was the same, but I should have included the multi namespace output.

zengin:
@MIchaelMainer, figured out the mutual dependency between PHP and PHPBeta tests. When they are run as part of the same process, `ConfigurationService.Settings` will refer to whichever runs first i.e. if PHP goes first PHPBeta doesn't get the necessary Beta prefix, if PHPBeta goes first, Beta prefix appears for PHP run as well.

PR #274, opened 2020-08-20
Bugfix/java import keyword

baywet:
fixes a bug where action/methods for have a reserved name in java


baywet:
you can see the result, on the beta codegen, here https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/5/commits/a8391ecb3b84245aa124f0c371412014d0dd8788

zengin:
This looks like a possible breaking change in V1 if we have those names there as well. What does the V1 diff look like?

baywet:
@zengin thanks for looking into it. I actually found other bugs generated by this specific method, this is why I reverted to draft for the time being.

I had a quick discussion with @darrelmiller about that earlier this morning. Basically there's no good solution here as the code is non deterministic and will sometimes do `msgraphReserved` and sometimes `msgraphreserved`.

I have a fix locally that will always do `msgraphReserved`, this will be a breaking change for existing properties, we'll simply have to bump major for v1 when it impacts it.

baywet:
@zengin just pushed my fix.
This code change makes the following changes in the generated metadata. https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/5/commits/2d83e4e11db947e886c22321c088ee74f5a6c725

I know it's a couple of breaking changes but at least now it's consistent.

zengin:
@baywet, I have nothing against breaking changes, just wanted to make sure that we mark this change as such so that we can batch multiple breaking changes in one release to have less major bumps in the version.

If V1 diff doesn't have any of these keywords, breaking beta shouldn't cause a major bump there. That's why I asked for V1 diff.

baywet:
@zengin I just ran it against v1, it does have a similar (smaller) breaking impact. Is there anything I should do to tag this PR as breaking?

baywet:
thanks! it's good for final review and merge then!

zengin:
I am hoping to release the namespace changes next week, and they may include breaking changes as well. Is this blocking you for Beta generation? If not, it would be great if we can batch this with namespace changes for a single major version bump.

MIchaelMainer:
Please note that a merge into dev will result in an automatic release for typewriter and will be used in the next pipeline generation. Unless you guys have objections, I'm going to update the generation pipeline to target `master`.

zengin:
@MIchaelMainer, no objections on my side.

MIchaelMainer:
Typewriter releases should now only happen on master.

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -10,17 +10,11 @@ namespace Microsoft.Graph.ODataTemplateWriter.CodeHelpers.Java
...
-        public static HashSet<string> ReservedNames
-        {
-            get
-            {
-                return new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
-                    "abstract", "continue", "for", "new", "switch", "assert", "default", "if", "package", "synchronized", "boolean", "do", "goto", "private", "this", "break", "double", "implements", "protected", "throw", "byte", "else", "import", "public", "throws", "case", "enum", "instanceof", "return", "transient", "catch", "extends", "int", "short", "try", "char", "final", "interface", "static", "void", "class", "finally", "long", "strictfp", "volatile", "const", "float", "native", "super", "while", "true", "false", "null"
-                };
-            }
-        }
+        public static Lazy<HashSet<string>> ReservedNames { get; private set; } = new Lazy<HashSet<string>>(() => new HashSet<string>(StringComparer.OrdinalIgnoreCase) {
```
zengin, line 15: Thanks for this perf improvement.

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -95,27 +89,19 @@ public static string GetToLowerFirstCharName(this OdcmProperty property)
...
             {
-                logger.Info("Property \"{0}\" is a reserved word in Java. Converting to \"{1}{0}\"", property, ReservedPrefix);
-                return ReservedPrefix + property;
+                var result = ReservedPrefix + property.ToUpperFirstChar();
+                logger.Info($"Property \"{property}\" is a reserved word in Java. Converting to \"{result}\"");
+                return result;
             }
-
-            if (odcmProperty != null && property == odcmProperty.Name.ToUpperFirstChar())
+            else if (property == odcmProperty?.Name?.ToUpperFirstChar() && !(odcmProperty?.Name?.StartsWith("_") ?? false))
```
zengin, line 98: no need for `else`

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -95,27 +89,19 @@ public static string GetToLowerFirstCharName(this OdcmProperty property)
...
-                    return string.Concat(property, "Property");
-                }
-
                 // Name the property by its type. Sanitize it in case the type is a reserved name.  
-                return odcmProperty.Projection.Type.Name.ToUpperFirstChar().SanitizePropertyName(odcmProperty);
+                return odcmProperty?.Projection?.Type?.Name?.ToUpperFirstChar()?.SanitizePropertyName(odcmProperty) ?? odcmProperty?.Name?.SanitizePropertyName();
             }
-
-            return property.Replace("@", string.Empty).Replace(".", string.Empty);
+            else
```
zengin, line 103: same here, no need for `else`

Templates/Java/BaseJavaModel.template.tt
```diff
@@ -1089,7 +1089,7 @@ public {1} {2}{3}{4} {{";
                 propertyName.SplitCamelCase(),
                 property.Name,
                 propertyType,
-                property.Name.SanitizePropertyName(property).ToLowerFirstChar(),
+                property.Name.ToLowerFirstChar().SanitizePropertyName(property),
```
zengin, line 1092: I think `ToLowerFirstChar()` is a no-op as all properties are camelCase.

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
baywet: yeah it called my attention....

Templates/Java/BaseJavaModel.template.tt
baywet: they should be, until you find out that some properties (under the device entity) are not camelCased and it breaks the generation.

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
baywet: it's a preference thing. I know it's a bit more verbose but it eases up my reading as I don't need to go into the body of the previous branch to understand only one of the branches will execute.

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
baywet: same comment as above.

baywet:
@zengin it's partially blocking, but I can merge my local generation and block any PR that reverts the changes. Anyway with @MIchaelMainer 's changes I'll merge that and we'll release with namespace changes

Templates/Java/BaseJavaModel.template.tt
zengin: interesting. Another example as to why we should have a validation tool for metadata for our assumptions in the code generation.

baywet: yep, here is my PR which attempts to fix the metadata in case you're interested in knowing which properties. https://microsoftgraph.visualstudio.com/onboarding/_git/AGS-OnboardingAutomationPipeline/pullrequest/5283

zengin: I wonder if it is a rule to have them in camelCase or me doing a hasty generalization :)

PR #278, opened 2020-08-26
java selective imports + a bunch of bugfixes + namespace fixes

baywet:
- fixes interfaces and types names for stream properties that should have included property name to avoid conflicts (breaking)
- adds selective imports to reduce build time and make building beta possible
- fixes Edm.Decimal which was not mapped at all
- fixes a bug where enums could have duplicate values
- fixes a bug where navigation properties would be missing
- fixes a bug where conflicting overfloads would be generated
- fixes a bug where BaseClient would do invalid and unecessary casting
- fixes an issues where entities ending in "Request" (search) would wreck the whole generation process
- fixes an issue where request and builders would generate duplicate methods
- fixes an issue where the documented type would be wrong
- moves utility import generation method into page class
- fixes an issue where interfaces would generate too many imports
- adds a method to get imports for method parameters
- adds casting of ctor parameters in call to disambiguous
- multiple fixes to namespace and types generation methods
- refactors type prefixing method for clarity
- removes star imports

I still need to do a few things (hence the draft status)
- [x] merge OdcmMethodReturnNamespaceSection into GetPackagePrefix
- [x] refactor CreatePackageDef OdcmClass case
- ~~remove last .*; from templates (2 left)~~


Templates/Java/requests_extensions/BaseEntityCollectionReferenceRequest.java.tt
```diff
@@ -6,6 +6,12 @@
 <#=writer.WriteHeader()#>
 <#=host.CreatePackageDef()#>
 
+import <#=host.CurrentModel.GetNamespace().AddPrefix()#>.options.QueryOption;
```
zengin, line 9: perf, global comment: when we have `host.CurrentModel.GetNamespace().AddPrefix()` more than once in a single file, can we replace that with a local variable instantiated once per file?

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -753,6 +752,29 @@ public static string CreatePackageDefForEnum(this CustomT4Host host)
...
+                return GetPackagePrefix((obj as OdcmParameter).Type);
+            }
+            else if (obj is OdcmProperty)
+            {
+                return GetPackagePrefix((obj as OdcmProperty).Type);
+            }
+            else
+            {
+                throw new ArgumentException("obj");
+            }
```
zengin, line 776: nit: `is` -> `as` can be bypassed by:
```suggestion
            switch(obj)
            {
                case OdcmEnum e:
                    return "";
                case OdcmType t:
                    return GetPrefixForModels();
                case OdcmParameter p:
                    return GetPackagePrefix(p.Type);
                case OdcmProperty p:
                    return GetPackagePrefix(p.Type);
                default:
                    throw new ArgumentException(nameof(obj));
            }
```

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -1165,15 +1185,35 @@ public static string CreatePackageDef(this CustomT4Host host)
...
+                case OdcmClass c:
+                    if (c.GetTypeString() != graphServiceEntityName)
+                        methodImports.Add(string.Format(importFormat, c.Namespace.Name.AddPrefix(), c.GetPackagePrefix(), c.GetTypeString()));
+
+                    var importTypeToExclude = host.TemplateFile.EndsWith("BaseEntityRequest.java.tt") ? host.TemplateName : string.Empty;
+                    c?.MethodsAndOverloads()?.Distinct()?.SelectMany(o => ImportClassesOfMethodParameters(o, importTypeToExclude: importTypeToExclude))?.ToList()?.ForEach(x => methodImports.Add(x));
+                    c?.NavigationProperties()?.Where(x => x.IsCollection)?.Select(x => x.Projection.Type)?.Distinct()?.ToList()?.ForEach(x =>
+                        ImportRequestBuilderTypes(host, x, methodImports, importFormat, interfaceTemplatePrefix, true)
+                    );
+                    c?.NavigationProperties()?.Where(x => !x.IsCollection)?.Select(x => x.Projection.Type)?.Distinct()?.ToList()?.ForEach(x =>
```
zengin, line 1214: nit: style: I find reading layered linq queries more readable when each function is in a different line (feels like SQL queries that way).

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -1165,15 +1185,35 @@ public static string CreatePackageDef(this CustomT4Host host)
...
+                    p.Projection?.Type?.AsOdcmClass()?.MethodsAndOverloads()?.Distinct()?.SelectMany(o => ImportClassesOfMethodParameters(o))?.ToList()?.ForEach(x => methodImports.Add(x));
                     break;
                 case OdcmMethod m:
-                    var sb = new StringBuilder(Environment.NewLine);
-                    methodImports = ImportClassesOfMethodParameters(m, "import {0}.{1}.{2};", sb).ToString();
+                    m.WithDistinctOverloads().SelectMany(o => ImportClassesOfMethodParameters(o))?.ToList()?.ForEach(x => methodImports.Add(x));
+                    goto default;
+                case OdcmClass c:
+                    if (c.GetTypeString() != graphServiceEntityName)
+                        methodImports.Add(string.Format(importFormat, c.Namespace.Name.AddPrefix(), c.GetPackagePrefix(), c.GetTypeString()));
```
zengin, line 1207: style: do you mind having curly brackets for single-line blocks as well to keep consistency in the code base? There are a few instances of this style but it is mostly styled as blocks.

baywet: I personally find it makes the syntax heavy to add curlys for single line blocks. Indentation should suffice from a readability perspective. Then again if it's a standard across the team, happy to oblige.

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
baywet: fixed

Templates/Java/requests_extensions/BaseEntityCollectionReferenceRequest.java.tt
baywet: fixed

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
zengin: I have seen bugs originating from someone forgetting to turn it into a block while adding another statement and somehow making its way into production, so it is also a safeguard measure for that potential error to be missed in code reviews.

+1 on less syntax, python is one of my favorite languages in terms of readability.

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -753,6 +752,22 @@ public static string CreatePackageDefForEnum(this CustomT4Host host)
             sb.Append("\n");
             return sb.ToString();
         }
+        public static string GetPackagePrefix(this OdcmObject obj)
+        {
+            switch(obj)
+            {
+                case OdcmEnum e:
+                    return "";
```
zengin, line 760: I made a typo in my suggestion
```suggestion
                    return "models.generated";
```

PR #281, opened 2020-09-03
- fixes a bug where url options would be inserted twice for collection POST requests

baywet:
## Summary

Because getRequestUrl already inserts the options (functions and query), providing the options again in build request was inserting those twice, causing the request to fail.

## Generated code differences

https://github.com/microsoftgraph/msgraph-sdk-java/pull/445

Provide the command line arguments here so that reviewers can quickly repro the results of changes.


zengin:
I understand the removal of `getOptions()`, but I am trying to understand addition of `getHeaders()`.

My assumption looking at the code change is that `getOptions()` was adding options as well as headers, and as you replace it with `getHeaders()` we are only keeping headers. Is that correct?

baywet:
yes we're only keeping the request headers. That's because getrequesturl already adds the functions and query parameters in the URL but not the headers.

zengin (approved):
Thanks for the clarification.

PR #282, opened 2020-09-03
- adds missing put methods for java sdk

baywet:
## Summary

the java SDK was missing put methods for entity requests. This PR fixes it.

## Generated code differences

https://github.com/microsoftgraph/msgraph-sdk-java/pull/446

Provide the command line arguments here so that reviewers can quickly repro the results of changes.

### DIFF
--- a/Templates/Java/requests_extensions/BaseEntityRequest.java.tt
+++ b/Templates/Java/requests_extensions/BaseEntityRequest.java.tt
@@ -50,6 +50,7 @@ import <#=importNamespace#>.http.HttpMethod;
 <#=deleteMethods(c)#>
 <#=patchMethods(c)#>
 <#=postMethods(c)#>
+<#=putMethods(c)#>
 <#  }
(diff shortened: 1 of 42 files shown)


zengin:
Do we need any applicability check to decide whether we have `PUT` available or not? It seems like the .NET version generates `PutAsync()` only on references. I am trying to understand if .NET needs fixing or if we need a guard in Java to generate only for references.

baywet:
we had a quit chat about that with @darrelmiller this morning.
As far as we know, there's no official annotation to indicate support or non support of the method.
We could inject our own custom annotation via XSLT, and then code gen the methods based on that annotation to make it more precise.
Going through all the entities to check which ones support PUT or not is going to be a painful exercise though.

baywet:
the other factor is that Java doesn't support partial classes or extensions methods, making it difficult to make a "one off" addition to what's codegen

zengin (approved):

PR #283, opened 2020-09-04
Add [Obsolete] attribute to deprecated CSharp classes

mikepizzo:
## Summary

Adds the generation of the [Obsolete] attribute when generating C# classes

## Generated code differences

Entity Types, Complex Types, Enum Types, EntitySets, Singletons, Functions and Actions marked as Deprecated in CSDL are annotated with the [Obsolete] attribute in C#.

## NOTE
Requires updated Vipr library to expose Deprecation on OdcmObject.


Templates/CSharp/Model/EnumType.cs.tt
```diff
@@ -35,6 +39,9 @@ namespace <#=enumT.Namespace.GetNamespaceName()#>
         /// <summary>
         /// <#=emumMember.Name.ToUpperFirstChar().SplitCamelCase()#>
         /// </summary>
+<# if (emumMember.Deprecation != null) { #>
+        <#=string.Format("[Obsolete(\"{0}\")]", emumMember.Deprecation.Description)#>
```
baywet, line 43: what's the benefit of using a string format here? (x3) IMHO it makes reading the template harder.

MIchaelMainer:
TODO:
- [ ] Update vipr submodule
- [ ] Support IMethodRequest, *StreamRequest, *EntityWithReferenceRequeest, *EntityRequest, *EntityReferenceRequest, *EntityCollectionWithReferencesRequest, *EntityCollectionRequest, *EntityCollectionReferencesRequest, MethodRequestBody.
- [ ] Create backlog items for supporting Java, Objective-C, TypeScript, and PHP.

baywet:
@MIchaelMainer the submodule will be updated automatically by a dependenbot pull request :)

MIchaelMainer:
@baywet Thank you. Yeah, these were notes for myself.

baywet:
FYi I just merged dev into this branch, which contains the submodule update. This way CI can run and validate this PR works.

mikepizzo:
Thanks @baywet! -- I see that the checks all now pass.

I saw that you had commented on the use of String.Format rather than just doing string concatenations in a few places.  I'm happy to fix that in places where it's a simple concatenation -- I was also thinking of defining  "[Obsolete(/"{0}/")]" as a string constant and then continuing to use String.Format using that constant, rather than duplicating the string in multiple places.  Happy to go either way -- just let me know which you prefer(and, if I go with the string constant approach, where I should define it).

Michael has created the related issue to track adding to the other templates.

I'd love to get this merged this week, if possible, before an SLT review on Friday.  What work is left in order to get it merged?  I see that @MIchaelMainer also suggested other places where we could add the obsolete attribute, but I don't know the code well enough to know in what unique scenarios these would appear.  My changes expose the [Obsolete] attribute for type definitions, their properties/navigation properties, methods, and request builders exposed through EntityClient. What other surface areas expose deprecated types/properties/methods that wouldn't be covered by one of the above? Can we add additional annotations later, as appropriate?

MIchaelMainer:
@mikepizzo That's fine. I've made the other template work a separate work item. You got the models. 
https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/356

Regarding the string format. Create an extension method on OdcmEnumMember (I don't know the type name) that will return the correct string. OdcmEnumMember.GetDeprecationString() or something like that.
https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/dev/src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs

mikepizzo:
@MIchaelMainer  -- Thanks for the info.

The string returned by GetDeprecationString() will be specific to the language, where-as it looks like OdcmModelExtensions is common across languages?  Would TypeHelperCSharp be a more appropriate place to put the method?

MIchaelMainer:
@mikepizzo Yes, TypeHelperCSharp is much more appropriate. Since it is likely an extension method on an Odcm type, let's name it something like GetCSharpDeprecationString()

mikepizzo:
@MIchaelMainer  -- shoot; I just pushed.

Looking at the code, I like your original suggestion of GetDeprecationDescription() since it's in the TypeHelperCSharp class (in the Microsoft.Graph.ODataTemplateWriter.CodeHelpers.CSharp namespace) so it's already scoped to CSharp. This is consistent with the other extension methods in the TypeHelperCSharp class that don't prefix with "CSharp" (for example, GetTypeString(), which is an extension method on OdcmType to return the C# type name).

So, I think your first suggestion is clearer and more consistent, but let me know if you want me to change it -- it looks like I have a template issue I need to fix anyway.

baywet (dismissed):

baywet:
@mikepizzo thanks for making the changes, much less duplication this way, it's going to be easier to maintain over time. I'm good with the current state of things. I'll let @MIchaelMainer and @zengin give it another review.

zengin:
The changes look good to me in the main code.

I have a couple of questions:
1. Why do we have test changes in this PR even though they don't correspond to the changes in the main code? Maybe it is a question to @baywet as the test commit belongs to him.
2. Could you add deprecation information into a couple of types in `\test\Typewriter.Test\Metadata\MetadataMultipleNamespaces.xml` and update the test files accordingly?

baywet:
@zengin I believe it's because the templates have been updated, if we don't have test data to for deprecation, a using System was added in all the templates, systematically. See [this change](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/283/files?file-filters%5B%5D=.tt#diff-3bfa17a630acc5dd06666ce4cb4ec3073a137e699d99de50300548df97475309)

MIchaelMainer:
+1 on @zengin 

> Could you add deprecation information into a couple of types in \test\Typewriter.Test\Metadata\MetadataMultipleNamespaces.xml and update the test files accordingly?

mikepizzo:
Added deprecation to a number of types in MetadataMultipleNamespaces.xml showing a variety of propagation scenarios, validated expected output for each case, and updated expected results as per @zengin. Please take a look.

@baywet  -- agree that putting the generation of the obsolete attribute in a common method was nice cleanup -- thanks for the suggestion @MIchaelMainer!

zengin (approved):
Thanks a lot for the tests.

MIchaelMainer (approved):

baywet (approved):

PR #288, opened 2020-09-10
Nikithauc/typescript beta mainnamespace

nikithauc:
For beta version, the typescript typings top-level namespace is 'microsoftgraphbeta'. The existing code has main namespaces always set to 'microsoftgraph'. This PR brings in the following changes-
1. For the main namespace to be generated as 'microsoftgraphbeta' when generating beta version, we will be passing "typescript.namespacePostfix:beta" as a property '-p typescript.namespacePostfix:beta'. 
2. During the generation of the typescript main namespace, the code will if the property 'typescript.namespacePostfix' exists and if present, will append the value to "microsoftgraph".

### DIFF
--- a/src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
+++ b/src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
@@ -43,7 +43,8 @@ namespace Microsoft.Graph.ODataTemplateWriter.CodeHelpers.TypeScript
         /// Returns the main or top level namespace
         /// </summary>
         /// <returns></returns>
-        public string GetMainNameSpace() {
+        public string GetMainNameSpace()
+        {
             return TypeScriptMainNamespaceName;
         }
 
--- a/src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
+++ b/src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.Graph.ODataTemplateWriter.CodeHelpers.CSharp;
+using Microsoft.Graph.ODataTemplateWriter.Settings;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -35,8 +36,16 @@ namespace Microsoft.Graph.ODataTemplateWriter.CodeHelpers.TypeScript
         // constants
         private const int MaxLineLength = 120;
         private const string MainNamespaceName = "Microsoft.Graph";
-        private const string TypeScriptMainNamespaceName = "microsoftgraph";
+        private string TypeScriptMainNamespaceName = "microsoftgraph" + (ConfigurationService.Settings.Properties != null && ConfigurationService.Settings.Properties.ContainsKey("typescript.namespacePostfix") ? ConfigurationService.Settings.Properties["typescript.namespacePostfix"] : string.Empty);
         private const string TabSpace = "    ";
+
+        /// <summary>
+        /// Returns the main or top level namespace
+        /// </summary>
+        /// <returns></returns>
+        public string GetMainNameSpace() {
+            return TypeScriptMainNamespaceName;
+        }
 
         /// <summary>
         /// Groups entity, complex and enum types to be printed
--- a/src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
+++ b/src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
@@ -36,16 +36,16 @@ namespace Microsoft.Graph.ODataTemplateWriter.CodeHelpers.TypeScript
         // constants
         private const int MaxLineLength = 120;
         private const string MainNamespaceName = "Microsoft.Graph";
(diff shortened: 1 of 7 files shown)


src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
```diff
@@ -35,9 +36,18 @@ public class TypeScriptNamespace
...
         private const string MainNamespaceName = "Microsoft.Graph";
-        private const string TypeScriptMainNamespaceName = "microsoftgraph";
+        private string TypeScriptMainNamespaceName = "microsoftgraph" + (ConfigurationService.Settings.Properties != null && ConfigurationService.Settings.Properties.ContainsKey("typescript.namespacePostfix") ? ConfigurationService.Settings.Properties["typescript.namespacePostfix"] : string.Empty);
         private const string TabSpace = "    ";
 
+        /// <summary>
+        /// Returns the main or top level namespace
+        /// </summary>
+        /// <returns></returns>
+        public string GetMainNameSpace()
```
zengin, line 46: nit: namespace is considered a single word.
```suggestion
        public string GetMainNamespace()
```

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
```diff
@@ -35,9 +36,18 @@ public class TypeScriptNamespace
         // constants
         private const int MaxLineLength = 120;
         private const string MainNamespaceName = "Microsoft.Graph";
-        private const string TypeScriptMainNamespaceName = "microsoftgraph";
+        private string TypeScriptMainNamespaceName = "microsoftgraph" + (ConfigurationService.Settings.Properties != null && ConfigurationService.Settings.Properties.ContainsKey("typescript.namespacePostfix") ? ConfigurationService.Settings.Properties["typescript.namespacePostfix"] : string.Empty);
```
zengin, line 39: This is no longer a constant. Consider calling this `TypeScriptMainNamespacePrefix` with a value `microsoftgraph` and move the beta appending logic into `GetMainNamespace()` function. Consider null-conditional operator for the null check as well.

zengin (changes requested):

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
nikithauc: Made the correction.

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
nikithauc: Done.

src/GraphODataTemplateWriter/CodeHelpers/TypeScript/TypeScriptNamespace.cs
```diff
@@ -36,16 +36,16 @@ public class TypeScriptNamespace
...
 
         /// <summary>
         /// Returns the main or top level namespace
         /// </summary>
         /// <returns></returns>
-        public string GetMainNameSpace()
+        public string GetMainNamespace()
         {
-            return TypeScriptMainNamespaceName;
+            return TypeScriptMainNamespaceNamePrefix + (ConfigurationService.Settings.Properties != null && ConfigurationService.Settings.Properties.ContainsKey("typescript.namespacePostfix") ? ConfigurationService.Settings.Properties["typescript.namespacePostfix"] : string.Empty);
```
zengin, line 48: style: Please consider splitting this into multiple lines. And if you use null-conditional operator:
```suggestion
            var properties = ConfigurationService.Settings?.Properties;
            return properties?.ContainsKey("typescript.namespacePostfix") == true
                ? TypeScriptMainNamespaceNamePrefix + properties["typescript.namespacePostfix"]
                : TypeScriptMainNamespaceNamePrefix;
```

zengin (approved):

test/Typewriter.Test/MultipleNamespacesTestRunner.cs
```diff
@@ -54,7 +54,7 @@ string getMetadataFile(TestLanguage testLanguage)
 
             // Arrange
             var languageStr = language.ToString();
-            var directoryPostfix = isPhpBeta ? "PHPBeta" : languageStr;
+            var directoryPostfix = languageStr + (isBeta ? "Beta" : string.Empty);
```
ddyett, line 57: not sure if case sensitivity matters here but for php would languageStr be "PHP" so the string ends up exactly the same. Since it's a passed in parameter i don't know that's guaranteed and if it matters.

ddyett: i think this is answered looking at how it's called from the PHPMultipleNamespaces file.

ddyett (approved):

test/Typewriter.Test/MultipleNamespacesTestRunner.cs
zengin: This is fed from an enum that is local to the tests, so nobody will be able to change it by changing a string parameter somewhere. The test caller needs to specify an enum value.

Also this is only used as part of a directory name where case doesn't matter as long as you are on Windows :)

PR #292, opened 2020-09-16
- adds a vararg support for build request in java templates

baywet:
## Summary

- adds vararg support for options in buildrequest in java template to allow consumers to provide options in a less verbose approach.
- adds missing docs on methods

## Generated code differences

https://github.com/microsoftgraph/msgraph-sdk-java/pull/477


Templates/Java/requests_extensions/BaseEntityCollectionReferenceRequestBuilder.java.tt
```diff
@@ -23,10 +24,22 @@ import <#=importNamespace#>.core.IBaseClient;
...
-    public <#=c.ITypeCollectionReferenceRequest()#> buildRequest() {
-        return buildRequest(getOptions());
+    /**
+     * Creates the request
+     *
+     * @param requestOptions the options for this request
+     * @return the IUserRequest instance
+     */
+    public <#=c.ITypeCollectionReferenceRequest()#> buildRequest(final com.microsoft.graph.options.Option... requestOptions) {
+        return buildRequest(requestOptions != null && requestOptions.length > 0 ? Arrays.asList(requestOptions) : getOptions());
```
zengin, line 34: Looks simple and unlikely to change, but if the check on requestOptions ever changes, it needs to be changed everywhere. This can be refactored in one of the 3 ways:
1. Wrapping it with a function that checks `requestOptions` and calls `getOptions()`
2. Pushing the logic down to `getOptions()`. This may break the semantics if it is currently a simple property access or of a similar nature.
3. Having a template variable for the body as the full line doesn't depend on any types and can be single string for T4 template.

1 & 2 may have a minor improvement on the binary size as well.

zengin (changes requested):

Templates/Java/requests_extensions/BaseEntityCollectionReferenceRequestBuilder.java.tt
baywet: thanks! updated

zengin (approved):

PR #295, opened 2020-09-18
feature/hashmap import java removal

baywet:
## Summary

removes unnecessary imports.

## Generated code differences

https://github.com/microsoftgraph/msgraph-sdk-java/pull/482


Templates/Java/models_extensions/BaseEntity.java.tt
```diff
@@ -7,11 +7,11 @@
 <#=host.CreatePackageDefForEntity()#>
 
 import com.google.gson.JsonObject;
-import com.google.gson.JsonElement;
 import com.google.gson.annotations.SerializedName;
 import com.google.gson.annotations.Expose;
+<# if(((OdcmClass)c).IsOpen && !((OdcmClass)c).Properties.Any()) { #>
```
zengin, line 12: nit: Looks like this is needed for AdditionalData. Consider leaving a comment on the relation between the import of HashMap and the condition.

zengin (approved):

Templates/Java/models_extensions/BaseEntity.java.tt
baywet: it's not for additional data but for open type with no properties (that are effectively hashmaps), things like PlannerChecklistItems for instance

PR #298, opened 2020-09-23
feature/collection navigation props raw object

baywet:
## Summary

- simplifies the parsing of navigation collection properties by using the serializer that already implements the code for it.
- removes subsequently unused references

## Generated code differences

https://github.com/microsoftgraph/msgraph-sdk-java/pull/497


MIchaelMainer (approved):
Also reviewed the generated code files.

PR #301, opened 2020-09-25
feature/java orderby

baywet:
## Generated code differences

https://github.com/microsoftgraph/msgraph-sdk-java/pull/503

### DIFF
--- a/Templates/Java/requests_extensions/BaseEntityCollectionReferenceRequest.java.tt
+++ b/Templates/Java/requests_extensions/BaseEntityCollectionReferenceRequest.java.tt
@@ -81,6 +81,19 @@ import <#=mainNamespace#>.<#=c.GetPackagePrefix()#>.<#=c.TypeName()#>;
     }
 
 <# } #>
+<# if (c.GetFeatures().CanSort) { #>
+    /**
+     * Sets the order by clause for the request
+     *
+     * @param value the sort clause
+     * @return the updated request
+     */
+    public <#=c.ITypeCollectionReferenceRequest()#> orderBy(final String value) {
+        addQueryOption(new com.microsoft.graph.options.QueryOption("$orderby", value));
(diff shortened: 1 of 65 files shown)


zengin (commented):
@baywet Do you have an example where `orderBy` is used on a non-collection? I looked at couple of OData docs and all the samples I saw were on collections even though they don't say the clause is exclusive to collections, e.g. https://docs.microsoft.com/en-us/odata/concepts/queryoptions-overview#orderby

baywet:
good catch. I think you're referring to these two templates specifically?
https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/56a5b257af6c0188fa97db6ce08346819f4120aa/Templates/Java/requests_extensions/BaseEntityRequest.java.tt
https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/blob/56a5b257af6c0188fa97db6ce08346819f4120aa/Templates/Java/requests_extensions/BaseEntityWithReferenceRequest.java.tt

I think what happened is that when I introduced filter, I did a search for expand. And went a bit too far. 
And the same thing happened for orderBy but with filter this time. I think both odata query parameters functions should be removed from these two.

Did I miss anything else?

zengin:
I think we should remove `orderBy` from all these as they are not operating on collections:
```
Templates/Java/requests_extensions/BaseEntityReferenceRequest.java.tt 
Templates/Java/requests_extensions/BaseEntityRequest.java.tt 
Templates/Java/requests_extensions/BaseEntityWithReferenceRequest.java.tt 
Templates/Java/requests_extensions/BaseMethodBodyRequest.java.tt 
Templates/Java/requests_extensions/BaseMethodRequest.java.tt 
```

baywet:
thanks! updated, sorry about the mess.

zengin (approved):

PR #303, opened 2020-09-29
feature/odata cast derivedconstraint

baywet:
## Summary

Adds support for odata casting in Java (and more generally in typewriter) by leveraging drived type constraints annotations.
replaces #302 

## links

https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Validation.V1.xml#L147
https://docs.microsoft.com/en-us/odata/webapi/built-in-routing-conventions

## Generated code differences

https://github.com/microsoftgraph/msgraph-sdk-java/pull/509


Templates/Java/requests_extensions/BaseEntityCollectionWithReferencesRequest.java.tt
```diff
@@ -45,44 +45,44 @@ import <#=importNamespace#>.concurrency.IExecutors;
     }
 
 <# if (c.GetFeatures().CanExpand) { #>
-    public <#=c.ITypeCollectionRequest()#> expand(final String value) {
+    public <#=c.ITypeCollectionWithReferencesRequest()#> expand(final String value) {
```
zengin, line 48: Looks like a breaking change.

Templates/Java/requests_extensions/BaseEntityCollectionWithReferencesRequest.java.tt
```diff
@@ -45,44 +45,44 @@ import <#=importNamespace#>.concurrency.IExecutors;
...
         addQueryOption(new com.microsoft.graph.options.QueryOption("$expand", value));
-        return (<#=c.TypeCollectionRequest()#>)this;
+        return this;
```
zengin, line 50: Q: Is this getting removed because it was no-op or is there another reason?

Templates/Java/requests_extensions/IBaseEntityCollectionRequest.java.tt
```diff
@@ -87,6 +87,6 @@ import <#=importNamespace#>.http.IHttpRequest;
      *
 	 * @return the updated request
 	 */
-	<#=c.ITypeCollectionRequest()#> skipToken(String skipToken);
+	<#=c.ITypeCollectionRequest()#> skipToken(final String skipToken);
```
zengin, line 90: You are a big advocate of compile-time constraints and I enjoy all these `final` changes. Really appreciate it!

zengin (changes requested):
E2E logic makes sense.

Waiting on:
- [ ] Are return type changes breaking?
- [ ] Decision from https://github.com/microsoft/Vipr/pull/147#discussion_r498552287, in case it requires changes in this repository.

Templates/Java/requests_extensions/BaseEntityCollectionWithReferencesRequest.java.tt
baywet: I removed it because it looked like an error and I don't think it ever worked to begin with (bad copy/paste)

TypeCollectionRequest will give us something like "GroupCollectionRequest" but `this` is a "GroupCollectionWithReferencesRequest". Even though they share some of the same methods, their first common ancestor is "BaseCollectionRequest".
On top of that the return type of the method is "ITypeCollectionWithReferencesRequest" which would give us "IGroupCollectionWithReferencesRequest", which is what the current class is implementing

Templates/Java/requests_extensions/IBaseEntityCollectionRequest.java.tt
baywet: the more work I can give to the compiler checking for what clumsy me can do, the better!

Templates/Java/requests_extensions/BaseEntityCollectionWithReferencesRequest.java.tt
baywet: on paper it is. I think that:
1. the cast never properly worked, either returning an invalid cast exception or up casting to the first common ancestor (BaseCollectionRequest).
2. the two types share a lot of methods in common, so even if somebody is referencing something, they should get a transparent switch in most cases.

baywet:
@zengin thanks for the great feedback! replied to both.

Templates/Java/requests_extensions/BaseEntityCollectionWithReferencesRequest.java.tt
zengin: Since it is already broken, I agree that we don't need to tag it as a breaking change.

Templates/Java/requests_extensions/BaseEntityCollectionWithReferencesRequest.java.tt
zengin: makes sense.

MIchaelMainer (approved):

zengin (changes requested):
- [ ] tests need the same updates

baywet:
> * [ ]  tests need the same updates

yep, I just saw that, on it :)

baywet:
@zengin I actually had forgotten a file in my last commit 🤦‍♂️

zengin (approved):

Issue #242, opened 2020-05-05
Remove scenarios where we set odata.type by default for C#


MIchaelMainer:
Only add Odata.type when working with derived types and heterogenous collections.

I think this issue is a duplicate 
[AB#7083](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/7083)

andrueastman:
Closed via #534

PR #312, opened 2020-10-16
Fix the setting of odata.type in the constructor of entitytype and complextype

MIchaelMainer:
## Summary

We only want to set odata.type under the following conditions:
1. a type is a derived type, and
2. one of its base types is referenced as the type for a property in another type. The odata.type is necessary for the derivedtypeconverter to properly deserialize the type. Current behavior is that we set odata.type on all entities and complex types. Some service APIs do not properly handle the unexpected instance annotation, even though they should ignore it.

This change removes the setting of odata.type on entities and complex types that do not meet the two conditions above.

## Todo

- [x] [Fix the hack](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/compare/bugfix/odataTypeCstor?expand=1#diff-d66e34ef80d710f4a6a2cdf93617b21d247a89e1fa2f7587a94e90af0136f3c1R348) where we exclude the OfficeGraph entity navigation properties that are of type entity. These overly broad references to a super base type make it impossible to determine which base types can be referenced. We need a better fix than the inline hack.

```xml
      <EntityType Name="sharedInsight" BaseType="graph.entity">
        <NavigationProperty Name="lastSharedMethod" Type="graph.entity" />
        <NavigationProperty Name="resource" Type="graph.entity" />
      </EntityType>
```

## Generated code differences

**Entity**
image

**Complex type**
image

`-v Info -m https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsv1.0.xml -o dotnetv1.0_fixOdataType2_with_complex -l CSharp -g Files`

Fixes https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/issues/242
Closes https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/67
Closes https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/67
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/598
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/566
Closes https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/560

### DIFF
--- a/Templates/templates/CSharp/Base/CollectionRequest.Base.template.tt
+++ b/Templates/templates/CSharp/Base/CollectionRequest.Base.template.tt
@@ -111,14 +111,6 @@ public string GetPostAsyncMethod(OdcmProperty odcmProperty, string @namespace, s
         stringBuilder.Append(Environment.NewLine);
         stringBuilder.AppendFormat("            this.Method = {0};", templateWriter.PostMethod);
 
-        var entity = odcmProperty.Projection.Type.AsOdcmClass();
-
-        if (entity != null && entity.IsAbstract)
-        {
-            stringBuilder.Append(Environment.NewLine);
-            stringBuilder.AppendFormat("            {0}.ODataType = string.Concat(\"#\", StringHelper.ConvertTypeToLowerCamelCase({0}.GetType().FullName));", sanitizedPropertyName);
-        }
-
         stringBuilder.Append(Environment.NewLine);
         if (isGraphResponseMethod)
         {
--- a/Templates/templates/CSharp/Model/ComplexType.cs.tt
+++ b/Templates/templates/CSharp/Model/ComplexType.cs.tt
@@ -48,6 +48,8 @@ if (complex.IsDeprecated)
    attributeStringBuilder.Append(complex.GetDeprecationString());
 }
 
+// We'll use this to generate a default OData.type only if the base is abstract or if its base is possibly referenced as a type in a property.
+var isBaseTypeReferenced = complex.IsBaseTypeReferenced() || (complex.Base != null && complex.Base.IsAbstract);
 
 // We only want to add the derived type converter to the class that has any derived types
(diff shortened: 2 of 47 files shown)


MIchaelMainer:
We've concluded that we won't make this change. 

We currently check the type and set the derived type when making requests that send a single item. For example, [adding an attachment to a message already in the mailbox](https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/blob/master/src/Microsoft.Graph/Generated/requests/MessageAttachmentsCollectionRequest.cs#L57) if the item is a FileAttachment. This won't work when the attachment is an ItemAttachment since we don't inspect the nested  types. So, for [adding an ItemAttachment](docs.microsoft.com/en-us/graph/api/message-post-attachments?view=graph-rest-1.0&tabs=http#example-item-attachment), the top level ODataType is for ItemAttachment, but the nested event type won't be set. Now, we could  inspect via reflection the payload for all types being sent and set the odatatype, but that could be 1) costly if the payload is large, 2) we may be back to the original situation where we are setting ODataType on objects for workloads that are not expecting the @odata.type annotation. This also affects [sendmail with attachments](https://docs.microsoft.com/en-us/graph/api/user-sendmail?view=graph-rest-1.0&tabs=http#request-3).

The current proposed fix is insufficient as any reference to an uber base type like entity causes all descendant items to have their odata.type set in the constructor, which brings us back to our original issue. 

We could remove generation of the odata.type, but that would now be a breaking change. 

We could create an annotation with a custom term to turn off scenarios, but that is again reactive to workloads not accepting odata.type. Ultimately, workloads should ignore instance annotations they don't understand, especially the common @odata.type instance annotation.

MIchaelMainer:
@andrueastman I'd like to discuss this with you.

andrueastman:
Closing this in favor of new PR

PR #332, opened 2020-11-23
feature/dotnet core

baywet:
This PR drafts the work to convert typewriter to dotnet core

- migrated to package references
- converts GraphODataTemplateWriter to dotnet core
- converts templates project to dotnet core
- converts template writer test project to dotnet core
- converts typewriter project to dotnet core
- converts typewriter test project to dotnet core
- removes additional feed reference
- updates reference to asynex
- updates reference to vipr
- migrates to roselyn compiler for templates compilation
- uses a temporary submodule for api doctor core (and markdowndeep core)
- revamps the configuration logic to make use of dotnet configuration mechanisms instead of deprecated library.

depends on:
- https://github.com/OneDrive/apidoctor/pull/54
- https://github.com/microsoft/Vipr/pull/164

I'm pausing at this point to allow for discussion but here are the main remaining changes required:
- inflector pluralizes type names, this is something we should be able to replace with another library or build ourselves. Although this gives warnings, it seems to be fine as this library is only using dotnet and no other dependency.


test/Typewriter.Test/Typewriter.Test.csproj
```diff
@@ -79,43 +29,324 @@
...
+    </PackageReference>
+    <PackageReference Include="NUnit.ApplicationDomain">
+      <Version>11.1.0</Version>
+    </PackageReference>
+    <PackageReference Include="NUnit3TestAdapter">
+      <Version>3.17.0</Version>
+    </PackageReference>
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Remove="TestDataCSharp\com\Microsoft\Graph2\CallRecords\Model\CallRecord.cs" />
```
zengin, line 62: I assume this allows file name patterns, so we can use `TestDataCSharp\**\*.cs`

src/GraphODataTemplateWriter/Settings/ConfigurationService.cs
```diff
@@ -66,39 +66,12 @@ private static TemplateWriterSettings LoadSettingsForLanguage()
...
-            ctorIL.Emit(OpCodes.Call, typeof(TemplateWriterSettings).GetMethod("CopyPropertiesFromMainSettings"));
-            ctorIL.Emit(OpCodes.Ret);
-
-            Type targetLanguageType = tb.CreateType();
-
             //Call the generic GetConfiguration method with our new type.
-            return (TemplateWriterSettings)typeof(IConfigurationProvider)
-                                            .GetMethod("GetConfiguration")
-                                            .MakeGenericMethod(targetLanguageType)
-                                            .Invoke(_configurationProvider, new object[] { });
```
zengin, line 101: Yes! Thank you! You are ticking off boxes from my todo list w.r.t engineering debt!

src/Typewriter/ConfigurationProvider.cs
```diff
@@ -1,21 +1,50 @@
...
+            if (!string.IsNullOrWhiteSpace(Environment.CurrentDirectory))
+            {
+                builder.AddJsonFile(GetConfigurationFilePath("TemplateWriterSettings.json"), true);
+                builder.AddJsonFile(GetConfigurationFilePath($"{targetLanguage}Settings.json"), false);
+            }
+
+            Configuration = builder.AddEnvironmentVariables().Build();
+        }
+        private string GetConfigurationFilePath(string fileName)
+        { //when using dotnet run the Environment.CurrentDirectory is set to csproj directory, not the bin like when runing with Visual Studio
```
zengin, line 37: nit: typo
```suggestion
        { //when using dotnet run, the Environment.CurrentDirectory is set to csproj directory, not the bin like when running with Visual Studio
```

src/Typewriter/ConfigurationProvider.cs
```diff
@@ -1,21 +1,50 @@
...
-using Its.Configuration;
 using Vipr.Core;
 
 namespace Typewriter
 {
-    public class ConfigurationProvider : IConfigurationProvider
+    public class ConfigurationProvider : Vipr.Core.IConfigurationProvider
     {
+        private readonly IConfiguration Configuration;
+        private const string dotNetCorePathprefix = "../GraphODataTemplateWriter";
```
zengin, line 12: nit: camelCase
```suggestion
        private const string dotNetCorePathPrefix = "../GraphODataTemplateWriter";
```

src/Typewriter/ConfigurationProvider.cs
```diff
@@ -1,21 +1,50 @@
...
+
+            Configuration = builder.AddEnvironmentVariables().Build();
+        }
+        private string GetConfigurationFilePath(string fileName)
+        { //when using dotnet run the Environment.CurrentDirectory is set to csproj directory, not the bin like when runing with Visual Studio
+            var regularPath = Path.Combine(Environment.CurrentDirectory, configFolder, fileName);
+            if (File.Exists(regularPath))
+                return regularPath;
+            else
+                return Path.Combine(Environment.CurrentDirectory, dotNetCorePathprefix, configFolder, fileName);
```
zengin, line 42:
```suggestion
                return Path.Combine(Environment.CurrentDirectory, dotNetCorePathPrefix, configFolder, fileName);
```

test/Typewriter.Test/Typewriter.Test.csproj
baywet: fixed, thanks for the suggestion

zengin:
@baywet do the tests work for you locally?

baywet:
@zengin good catch! I meant to come back to fixing the tests but I somehow forgot to capture the task 10 days ago. I just pushed the fix.
QQ: is the ObjctiveC test disabled somehow on CI? it's the only one that still fails locally and looking at other PRs I don't see it being executed

src/GraphODataTemplateWriter/Settings/ConfigurationService.cs
```diff
@@ -74,11 +77,18 @@ private static TemplateWriterSettings LoadSettingsForLanguage()
             return mainTWS;
 
         }
+        /// <summary>
+        /// Cheap "caching" implementation that compromises bettween performance impact, avoiding flaky unit tests because of a shared app domain and not having to refactor this whole settings implementation
+        /// </summary>
+        private static Dictionary<string, TemplateWriterSettings> cheapSettingsCache = new Dictionary<string, TemplateWriterSettings>();
```
zengin, line 83: nit: I think we can drop "cheap" from the name.

btw, great workaround!

baywet: I forgot that cheap can have a negative connotation in English, what do you think of "simple"?

baywet:
@zengin @MIchaelMainer this is ready for review, the unit tests are finally fixed! See the [latest passing build](https://o365exchange.visualstudio.com/O365%20Sandbox/_build/results?buildId=4256862&view=results)

Once merged here is our To-do list:
- [ ] merge the draft build [1](https://o365exchange.visualstudio.com/O365%20Sandbox/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=10728) & [2](https://o365exchange.visualstudio.com/O365%20Sandbox/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=10727)
- [ ] trigger code-gen for all languages (to avoid issues, we replaced the relative path for the template name by just the template file name, this is going to generate huge diffs)
- [ ] potentially fix the exe path in [the run typewriter script](https://github.com/microsoftgraph/msgraph-metadata/blob/master/scripts/runTypewriter.ps1)

src/GraphODataTemplateWriter/Settings/ConfigurationService.cs
zengin: I didn't think about the negative connotation, I thought SettingsCache was descriptive enough as that is the cache we use for the settings at the moment.

What is the additional information we convey with "simple"?

baywet: that the intention is to keep it simple and this is why it's not implemented with a proper DP.

zengin: Usually, workarounds tend to stick around for longer than anticipated lifetime. It makes sense to take a note on the implementation detail, which is already covered by the comment and the implementation itself, but it increases the cognitive load slightly while reading all the other references:

- At this line I am checking the simple settings cache
- At this line I am adding to the simple settings cache

vs

- At this line I am checking the settings cache
- At this line I am adding to the settings cache

Variable naming is a very opinionated matter, so I wanted to share my reasoning. It continues to qualify as nitpick ;)

baywet: thanks for sharing the detailed reasoning.
I'm personally used to longer variable names when they are non-trivial to express the intent, which helps with differentiation and other aspects. But I agree that practice increases the "parsing effort" (it's more verbose).
Let's have @MIchaelMainer be the "innocent and impartial" third party on this one and help us reach quorum :)

zengin (approved):
LGTM.

I am OK with merging this to `dev` and running a full generation after tomorrow's releases for all the languages.

The only change that is parked at `dev` w.r.t `master` is deprecation, and that shouldn't generate any significant diffs (except for a few `using`s).

@MIchaelMainer, do you have any concerns regarding that release plan?

src/GraphODataTemplateWriter/Settings/ConfigurationService.cs
MIchaelMainer: Because 

> workarounds tend to stick around for longer than anticipated lifetime. 

We call it SettingsCache. If changes need to be done to it, we'll read the comments then. No need to pass around comment info in the variable name.

MIchaelMainer:
This works for me.

MIchaelMainer (approved):

baywet:
thanks team! merged. Let's move it to master after today's codegen.

PR #339, opened 2020-12-03
feature/java count

### LINKED ISSUE #318: odata count segment path methods are not generated for java
e.g. https://docs.microsoft.com/en-us/graph/api/group-list-transitivemembers?view=graph-rest-1.0&tabs=java

baywet:
fixes #318
- updates test files
- adds support for odata count in java generation

generation result https://github.com/microsoftgraph/msgraph-sdk-java/pull/584


nikithauc:
Is there a way to set a default boolean value  for count ? https://github.com/microsoftgraph/msgraph-sdk-javascript/issues/241#issuecomment-719145816

baywet:
@nikithauc if you are referring to using default parameters, [java doesn't support it](https://stackoverflow.com/a/997529/3808675). This is why I added an overload for the default value.

zengin (approved):

PR #365, opened 2020-12-17
- replaces inflector by its dotnet standard verssion

baywet:
inflector hasn't been upgraded to dotnet standard. However the community put together a fork on dotnet standard that's more popular than the original thing.

### DIFF
--- a/src/GraphODataTemplateWriter/CodeHelpers/PHP/TypeHelperPHP.cs
+++ b/src/GraphODataTemplateWriter/CodeHelpers/PHP/TypeHelperPHP.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+﻿// Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 
 namespace Microsoft.Graph.ODataTemplateWriter.CodeHelpers.PHP
 {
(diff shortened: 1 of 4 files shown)


MIchaelMainer (dismissed):
A fork of a fork. Did a quick walk through the commit history since the original fork.

zengin (approved):
> inflector hasn't been upgraded to dotnet standard. However the community put together a fork on dotnet standard that's more popular than the original thing.

Love this about open source!

baywet:
@MIchaelMainer @zengin I've cleaned up the static references and replaced by static instances with culture information. This is good for final review.

src/GraphODataTemplateWriter/CodeHelpers/PHP/TypeHelperPHP.cs
```diff
@@ -188,7 +188,7 @@ public static string GetPHPNamespace(this string @namespace, string prefix = "")
 
             pieces.AddRange(@namespace.Split('.').ToList());
 
-            var pascalCasePieces = pieces.Select(piece => piece.Pascalize());
+            var pascalCasePieces = pieces.Select(piece => piece.ToPascalize());
```
zengin, line 191: old name makes more sense :)

zengin (approved):

src/GraphODataTemplateWriter/CodeHelpers/PHP/TypeHelperPHP.cs
baywet: I agree, however old name maps directly to the inflector extension method when the "new" one maps to our own extension methods. This is the only place as far as I remember that was diverging. I think the original idea behind using our own extension method was to encapsulate inflector.

zengin: Yeah, we should probably rename our extension method (not in this PR) to something like: `ToPascalCase()`.

PR #368, opened 2020-12-17
feature/optional method parameters

### LINKED ISSUE #261: Support arbitrary argument lists for actions in Java
OData actions are not overloaded by parameter lists. This means that new parameters can be added to an action without creating a new action. Currently, we generate method signatures based on the parameter list. An update to an action means a new API signature is created and that can break clients developed against existing signatures. 

* maintain existing signatures - this means we need some sort of baseline to determine if there are changes to parameter lists. 
* add support for arbitrary parameter lists This should be an overload for our generated signatures.

This also adds a beneficial side effect that allows the reordering of parameters. 

cc: @darrelmiller  
[AB#6921](https://microsoftgraph.visualstudio.com/0985d294-5762-4bc2-a565-161ef349ca3e/_workitems/edit/6921)

baywet:
- cleans up settings from inexisting templates
- adds a method to get the body builder - adds methods to get types distinct methods by name - adds support for different types of methods comparisons
- bumps language support to latest
- adds body generation for functions
- remaps methods with body to main template for java
- adds generation of body builder as well as functions options extraction methods
- adds support for methods optional parameters
- removes unecessary template
- java method request templates cleanup
- renames body to parameter set and adds delta link parsing
- moves the deltaLink and deltaToken methods to the request from the builder
- updates test files

See the detailed explanation in https://github.com/microsoftgraph/msgraph-sdk-java/pull/602

Fixes #261


zengin (approved):
👍

PR #383, opened 2021-01-04
Add metadata capture and cleaning

MIchaelMainer:
- update documentation for update documentation usage.
- update stage conditions.

Validation below consists of running generation from the UI with `Skip metadata capture and clean` set to true. This will get us to a clean state in the repo with minor changes from templates.. We will want to run this again when we know there are metadata changes to validate that it works in that situation. At that point, we can merge this in.


.azure-pipelines/generation-pipeline.yml
```diff
@@ -196,8 +258,14 @@ stages:
 
 - stage: stage_php_v1
   dependsOn:
+  - stage_build_and_publish_typewriter
   - stage_v1_metadata
```
zengin, line 262: I assume this doesn't block if `stage_v1_metadata` is skipped, i.e. `condition` below overrules this.

scripts/download-diff-metadata.ps1
```diff
@@ -0,0 +1,104 @@
...
+# Are we on master? If not, we will want our changes committed on master.
+$branch = &git rev-parse --abbrev-ref HEAD
+Write-Host "Current branch: $branch"
+if ($branch -ne $env:targetBranch) {
+    git fetch origin $env:targetBranch | Write-Host
+    git checkout $env:targetBranch | Write-Host
+    $branch = &git rev-parse --abbrev-ref HEAD
+    Write-Host "Current branch: $branch"
+    git pull origin $env:targetBranch --ff-only | Write-Host
+}
```
zengin, line 50: Aren't we in control of what branch we are on now? Would there be any scenario where we have a feature branch, and want to push into that particular branch?

scripts/run-typewriter-clean-metadata.ps1
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+
+Write-Host "Running Typewriter to clean the metadata..."
+Push-Location -Path $env:TypewriterDirectory
+
+Write-Host "Typewriter.exe: $env:TypewriterExecutable `nInput metadata: $env:InputMetadataFile `n output path: $env:OutputPath"
```
zengin, line 6: Can we remove this given that I have the logger enabled in Typewriter.exe to print out the arguments now?

zengin (approved):

.azure-pipelines/generation-pipeline.yml
MIchaelMainer: Correct, the condition overrules and helps with flow control. nit: the diagramming tool uses the dependsOn too to show the relationships between stages. I bet that if we group stacked the stages in the yaml based on the metadata stage dependency it would show a lot better in the diagram.

scripts/download-diff-metadata.ps1
MIchaelMainer: We always target master for the metadata. ADO checks out the repo detached HEAD state. I don't think we need feature branches here. This is old so perhaps there is better way. Do we even need to checkout master? Can we add and commit the changes in a detached HEAD state and push to origin/master? Any guidance is appreciated.

MIchaelMainer: I think this can be changed to 

`git checkout master`

zengin: If Azure DevOps doesn't fetch all the branches, we may need to keep `git fetch origin master` before `git checkout master` as well,

zengin (approved):

PR #385, opened 2021-01-05
Represent Duration Object in PHP as DateInterval

revoltek-daniel:
## Summary

Use DateInterval as Duration Object in PHP

The Duration Objects value is like "P1D" which could be represented by an PHP DateInterval Object. The generated Duration Object is not found in the PHP Models, like in https://github.com/microsoftgraph/msgraph-sdk-php/blob/dev/src/Beta/Microsoft/Graph/Model/BookingSchedulingPolicy.php#L161

Documentation for DateInterval
https://www.php.net/manual/en/dateinterval.construct.php

## Generated code differences
for Example:
https://github.com/microsoftgraph/msgraph-sdk-php/blob/dev/src/Beta/Microsoft/Graph/Model/BookingSchedulingPolicy.php#L161

`$this->_propDict["timeSlotInterval"] = new Duration($this->_propDict["timeSlotInterval"]);`
should be changed to
`$this->_propDict["timeSlotInterval"] = new \DateInterval($this->_propDict["timeSlotInterval"]);`


zengin (dismissed):
Thanks for the contribution @revoltek-daniel!

@MIchaelMainer, this change looks good to me.

DateInterval has a [constructor](https://www.php.net/manual/en/dateinterval.construct.php)  that takes ISO 8601 string.

[edm:Duration](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_Toc453752639) is based on [W3C dayTimeDuration](https://www.w3.org/TR/xmlschema11-2/#dayTimeDuration), which is derived from [W3C duration](https://www.w3.org/TR/xmlschema11-2/#duration), and that follows ISO 8601.

We have the same standard in our [.NET library](https://github.com/microsoftgraph/msgraph-sdk-dotnet-core/blob/dev/src/Microsoft.Graph.Core/Models/Duration.cs#L31)

MIchaelMainer (dismissed):
We will need to bump major version to capture this breaking change. This will need to be documented in the release notes.

Let's merge this in after the beta doc annotation update so we can see where the generated Duration changes occur (we could also just search the metadata).

zengin:
@MIchaelMainer, given that there is no `Microsoft\Graph\Model\Duration` implementation, nobody could have taken a typed dependency on this though, right? I think we can take this as a "non-breaking" breaking change without a major version bump (i.e. it fixes broken functionality).

MIchaelMainer:
@zengin  You're correct! Wow, I'm surprised that we didn't create a Duration object for this scenario AND this never came up before (or more likely, we didn't see it). No major bump.

zengin:
@MIchaelMainer, yeah, one of the downsides of PHP being an interpreted language.

I have just installed a VSCode extension called Intelliphense and it is able to highlight errors like these. I think we should consider doing that or a similar analysis for all our models. I see Duration-like errors, e.g. `Microsoft\Graph\Model\Single` doesn't exist either.

nikithauc:
> @MIchaelMainer, yeah, one of the downsides of PHP being an interpreted language.
> 
> I have just installed a VSCode extension called Intelliphense and it is able to highlight errors like these. I think we should consider doing that or a similar analysis for all our models. I see Duration-like errors, e.g. `Microsoft\Graph\Model\Single` doesn't exist either.

Is there an issue for -> `Microsoft\Graph\Model\Single` doesn't exist either. ?

zengin:
@nikithauc, opened this: https://github.com/microsoftgraph/msgraph-sdk-php/issues/369

Ndiritu:
Changed base branch to `support/php-2.x.x` before realising this isn't a breaking change.
Sorry to request your review again @MIchaelMainer @zengin 
Any reason this wasn't merged before?

MIchaelMainer (approved):
No reason from me. I think this just fell off our radar.

zengin:
Same here. Let's merge 🚀

zengin (approved):

PR #397, opened 2021-01-13
replaces the callback API surface by futures for java

baywet:
- updates java templates to provide Future api surface instaead of callbacks
- removes unecessary replacements for deleted interfaces - updates repalcement for futures
- updates test files
- removes unecessary async modifier on non-awaited method to suppress warning
- removes the super syntax as it becomes unecessary
- updates test data
- switches to completable futures
- updates test files

related https://github.com/microsoftgraph/msgraph-sdk-java/pull/610


Templates/templates/Java/requests_extensions/BaseMethodRequest.java.tt
```diff
@@ -128,10 +117,11 @@ import <#=mainNamespace#>.<#=c.GetPackagePrefix()#>.<#=c.AsOdcmMethod().TypePara
     /**
      * Gets the <#=c.ReturnType()#>
      *
-     * @param callback the callback to be called after success or failure
+     * @return a future with the result
      */
-    public void get(@Nonnull final ICallback<? super <#=c.ReturnType()#>> callback) {
-        send(HttpMethod.GET, callback, null);
+    @Nonnull
```
MIchaelMainer, line 122: I think we should remove `@Nonnull` as there could be method responses that contain an empty object. If a method returns an empty object, or a null object. The default is nullable=true for return types. How does Java client handle a `{}` for response versus an empty response? 


<Action Name="invite" IsBound="true" xmlns="http://docs.oasis-open.org/odata/ns/edm">   <Parameter Name="bindingParameter" Type="Collection(graph.participant)" />   <Parameter Name="participants" Type="Collection(graph.invitationParticipantInfo)" Nullable="false" />   <Parameter Name="clientContext" Type="Edm.String" Unicode="false" />   <ReturnType Type="graph.inviteParticipantsOperation" /> </Action>
--
<Action Name="renewGroup" IsBound="true" xmlns="http://docs.oasis-open.org/odata/ns/edm">   <Parameter Name="bindingParameter" Type="Collection(graph.groupLifecyclePolicy)" Nullable="false" />   <Parameter Name="groupId" Type="Edm.String" Nullable="false" Unicode="false" />   <ReturnType Type="Edm.Boolean" Nullable="false" /> </Action>

Templates/templates/Java/requests_extensions/BaseEntityRequest.java.tt
```diff
@@ -230,10 +235,11 @@ import <#=importNamespace#>.http.HttpMethod;
      * Creates a {0} with a new object
      *
      * @param new{0} the new object to create
-     * @param callback the callback to be called after success or failure
+     * @return a future with the result
      */
-    public void post(@Nonnull final {0} new{0}, @Nonnull final ICallback<? super {0}> callback) {{
-        send(HttpMethod.POST, callback, new{0});
+    @Nonnull
+    public java.util.concurrent.CompletableFuture<{0}> postAsync(@Nonnull final {0} new{0}) {{
```
MIchaelMainer, line 241: There are POST scenarios that result in a 204 No Content so that content can be null. Overall, I'm hesitant to attribute with NotNull unless the metadata is explict.

Templates/templates/Java/requests_extensions/BaseMethodRequest.java.tt
baywet: Whatever the service responds, we'll still have a future being returned. (Or an exception thrown) the result of that future might be null but that's not what this method is returning directly

Templates/templates/Java/requests_extensions/BaseEntityRequest.java.tt
baywet: Same comment as above. Futures are the equivalent of task in dot net. It's not because that task doesn't return anything that you don't have a task

Templates/templates/Java/requests_extensions/BaseMethodRequest.java.tt
MIchaelMainer: Got it.

MIchaelMainer (approved):

PR #419, opened 2021-01-28
Update per package compliance

MIchaelMainer:


.azure-pipelines/generation-templates/objc.yml
```diff
@@ -5,3 +5,9 @@ steps:
     BuildConfiguration: $(buildConfiguration)
     OutputFullPath: $(typewriterDirectory)/output
     RepoModelsDir: $(Build.SourcesDirectory)/msgraph-sdk-objc-models/GeneratedModels/
+- pwsh: |
+      gem install xcodeproj
+      & chmod +x .\add_generated_models.rb
+      ruby .\add_generated_models.rb | Write-Host
```
zengin, line 11: forward slash for unix environment.
```suggestion
      & chmod +x ./add_generated_models.rb
      ruby ./add_generated_models.rb | Write-Host
```

MIchaelMainer: Thank you!. I have that change locally. I'm hitting a perm issue installing the gem.

.azure-pipelines/generation-templates/objc.yml
```diff
@@ -5,3 +5,9 @@ steps:
     BuildConfiguration: $(buildConfiguration)
     OutputFullPath: $(typewriterDirectory)/output
     RepoModelsDir: $(Build.SourcesDirectory)/msgraph-sdk-objc-models/GeneratedModels/
+- pwsh: |
+      gem install --user-install xcodeproj
+      & chmod +x ./add_generated_models.rb
```
zengin, line 10: Do we need this line? I think the executable in next step is `ruby` so I don't think we need to convert the script itself to be an executable.

zengin (dismissed):

zengin (approved):

nikithauc (approved):

PR #473, opened 2021-03-04
Entities without base types should implement JsonSerializable

MIchaelMainer:
## Summary

Print and IdentityGovernance entities do not have a base type. Generator was creating a class declaration like: `class IdentityGovernance extends` with no base type to extend. PHP EntityType.php.tt assumed only Entity has no base type. This update has all types without base type implementing JsonSerializable.

I don't know why this didn't block the beta generation. 

Note that the Print entity becomes the GraphPrint class as `print` is a PHP keyword.

Removed white space.

## Generated code differences

TODO


src/Typewriter/Properties/launchSettings.json
```diff
@@ -1,8 +0,0 @@
-{
```
baywet, line 1: maybe we should exclude the launch settings from the repo and provide templates to easily onboard new developers? And do the same for vscode settings
(for some reason I can't comment on the file below with the path)

baywet (changes requested):
I'm not sure why but it looks like php unit tests are failing

PR #480, opened 2021-03-12
Entities without base types should implement JsonSerializable

MIchaelMainer:
##Summary
Print and IdentityGovernance entities do not have a base type. Generator was creating a class declaration like: `class IdentityGovernance extends` with no base type to extend. PHP EntityType.php.tt assumed only Entity has no base type, which was wrong as there are singletons which don't require an id property. This update has all types without base type implementing JsonSerializable.

I don't know why this didn't block the beta generation.

Note that the Print entity becomes the GraphPrint class as print is a PHP keyword.

replace #473


zengin:
@MIchaelMainer the reason Beta is not blocked is that Beta generation has the old version of the V1 when checkout happens. If the tests are targeting only the V1 types, running tests will not cause any failure in test step for Beta.

zengin (approved):

PR #498, opened 2021-03-18
Add backslash to namespace references

SilasKenneth:
## Summary

Currently all `getXXX()` methods that make calls to class in other namespaces in the SDK fail to resolve the classes due to missing backslash at the beginning of the class instantiation, for instance,

```php
  function getXXX(){
        return new NameSP\ClassName($argument);
   }
``` 
should be
```php
  function getXXX(){
        return new \NameSP\ClassName($argument);
   }
``` 
Sample code with the issue.
https://github.com/microsoftgraph/msgraph-sdk-php/blob/5e4332208618c80b678db614edf25fc0e24e5b4b/src/CallRecords/Model/CallRecord.php#L164

## Generated code differences
Screenshot 2021-03-18 133834

`-v Info -m C:\\github\\msgraph-metadata\\clean_v10_metadata\\cleanMetadataWithDescriptionsv1.0.xml -g Files -l ObjC -o ObjC_v1.0 -e v1.0`

https://github.com/microsoftgraph/msgraph-sdk-php/issues/424
https://github.com/microsoftgraph/msgraph-sdk-php/pull/412


MIchaelMainer (commented):
Very nicely formed PR with repro and notes! You ran the tests and updated the test files - very good. Nice catch that this affects complexType as well. Also, you want to make sure that you ran this against 

To complete this, let's link to PRs in msgraph-sdk-php created by this change for both v1.0 and beta models. You can either run your changes locally or use the pipeline to run it. If you do use the pipeline, make sure that we skip metadata capture and clean so that we use the last metadata (in case new metadata has been published since the last time we captured it). And of course, we need  @Ndiritu to sign off as well.

test/Typewriter.Test/TestDataPHP/com/Microsoft/Graph/CallRecords/Model/CallRecord.php
```diff
@@ -59,7 +59,7 @@ public function setVersion($val)
     public function getType()
     {
         if (array_key_exists("type", $this->_propDict)) {
-            if (is_a($this->_propDict["type"], "Microsoft\Graph\CallRecords\Model\CallType")) {
+            if (is_a($this->_propDict["type"], "\Microsoft\Graph\CallRecords\Model\CallType")) {
```
Ndiritu, line 62: @MIchaelMainer just curious...

For `v1` models, why do we generate the CallRecords models in their own namespace outside the `\Microsoft\Graph\Model` namespace? 

Also noticed that for `beta` we have `Ediscovery` and `TermStore` outside `Beta\Microsoft\Graph\Model` namespace as well.

test/Typewriter.Test/TestDataPHP/com/Microsoft/Graph/CallRecords/Model/CallRecord.php
```diff
@@ -206,15 +206,15 @@ public function setEndDateTime($val)
     /**
     * Gets the organizer
     *
-    * @return Microsoft\Graph\Model\IdentitySet The organizer
+    * @return \Microsoft\Graph\Model\IdentitySet The organizer
     */
     public function getOrganizer()
     {
         if (array_key_exists("organizer", $this->_propDict)) {
```
Ndiritu, line 213: Should we also take this opportunity to factor in some changes from a community member ?(https://github.com/microsoftgraph/msgraph-sdk-php/pull/156) considering it affects many files

We can add the `$this->getProperties()` to the template

Ndiritu (approved):
Changes look good. Added a question about our model namespacing & consideration for throwing in a change proposed by a community member.

test/Typewriter.Test/TestDataPHP/com/Microsoft/Graph/CallRecords/Model/CallRecord.php
zengin: I shared the namespaces design document with you and Kenneth.

test/Typewriter.Test/TestDataPHP/com/Microsoft/Graph/CallRecords/Model/CallRecord.php
SilasKenneth: Hey @Ndiritu, thanks for the heads-up, will be making the necessary changes and do more improvements to the template file.

SilasKenneth:
> Very nicely formed PR with repro and notes! You ran the tests and updated the test files - very good. Nice catch that this affects complexType as well. Also, you want to make sure that you ran this against
> 
> To complete this, let's link to PRs in msgraph-sdk-php created by this change for both v1.0 and beta models. You can either run your changes locally or use the pipeline to run it. If you do use the pipeline, make sure that we skip metadata capture and clean so that we use the last metadata (in case new metadata has been published since the last time we captured it). And of course, we need @Ndiritu to sign off as well.

Thanks @MIchaelMainer, I will be running the pipelines, I had done that then I closed the PR(https://github.com/microsoftgraph/msgraph-sdk-php/pull/423) on the `msgraph-sdk-php`, will do it again.

test/Typewriter.Test/TestDataPHP/com/Microsoft/Graph/CallRecords/Model/CallRecord.php
zengin: I would recommend taking these fixes separately so that the commit history will be easy to understand for anyone looking back. It is even more useful when one of the fixes causes a bug as a side effect and if it needs to be reverted because of that.

I would recommend batching for breaking changes to release in a single major version, but even then having separate PRs helps a lot in the scenario described above.

SilasKenneth: Yes, that makes sense @zengin

MIchaelMainer (approved):

PR #512, opened 2021-04-12
Fix next link initialization

andrueastman:
This PR closes https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/275

It fixes the initialization of collection properties since the use of the `as operator keyword` used in the previous Newtonsoft compatible code cannot be used with System.Text.Json's [JsonElement](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement?view=net-5.0) since it is a struct and not nullable. This results to the nextlinks not being initialized.

Tests have been updated to capture this as well.

Generation PRs
V1 - https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/960
Beta - https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/pull/277
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/512)

### DIFF
--- a/Templates/templates/CSharp/Base/SharedCSharp.template.tt
+++ b/Templates/templates/CSharp/Base/SharedCSharp.template.tt
@@ -334,5 +334,40 @@ public string GetMethodRequestPrimitiveReturnTypeString(string type)
     }
 }
 
+/// <summary>
+/// Used in EntityCollectionResponse.cs.tt, EntityCollectionWithReferencesResponse and  MethodCollectionResponse.cs.tt
+/// to get the function to initialized the NextPageRequest object in the response of the collection request
+/// and to copy the Additional data to the CollectionPage instance
+/// </summary>
+/// <param name="iBaseClientTypeName"> type of the IBaseClient to use in creating the method</param>
+/// <returns>the string representation of the method</returns>
+public string GetInitializeCollectionPropertiesMethod(string iBaseClientTypeName)
+{
+    return @"/// <summary>
+        /// Initializes the NextPageRequest from the AdditionalData if possible
+        /// and copies the Additional data to the CollectionPage instance
+        /// </summary>
+        " + string.Format("internal void InitializeCollectionProperties({0} client)", iBaseClientTypeName)
+                + @"
+        {
+            if(AdditionalData != null)
+            {
+                if(AdditionalData.TryGetValue(""@odata.nextLink"", out var nextPageLink))
+                {
+                    // Ensure it is a non empty JsonElement string
+                    if (nextPageLink is System.Text.Json.JsonElement element
+                        && element.ValueKind == System.Text.Json.JsonValueKind.String
+                        && !string.IsNullOrEmpty(element.GetString()))
+                    {
+                        Value.InitializeNextPageRequest(
+                            client,
+                            element.GetString());
+                    }
+                }
+                // Copy the additional data collection to the page itself so that information is not lost
+                Value.AdditionalData = AdditionalData;
+            }
+        }";
+}
 
 #>
--- a/Templates/templates/CSharp/Base/SharedCSharp.template.tt
+++ b/Templates/templates/CSharp/Base/SharedCSharp.template.tt
@@ -334,40 +334,5 @@ public string GetMethodRequestPrimitiveReturnTypeString(string type)
     }
 }
 
-/// <summary>
-/// Used in EntityCollectionResponse.cs.tt, EntityCollectionWithReferencesResponse and  MethodCollectionResponse.cs.tt
-/// to get the function to initialized the NextPageRequest object in the response of the collection request
-/// and to copy the Additional data to the CollectionPage instance
-/// </summary>
-/// <param name="iBaseClientTypeName"> type of the IBaseClient to use in creating the method</param>
-/// <returns>the string representation of the method</returns>
-public string GetInitializeCollectionPropertiesMethod(string iBaseClientTypeName)
-{
-    return @"/// <summary>
-        /// Initializes the NextPageRequest from the AdditionalData if possible
-        /// and copies the Additional data to the CollectionPage instance
-        /// </summary>
-        " + string.Format("internal void InitializeCollectionProperties({0} client)", iBaseClientTypeName)
-                + @"
-        {
-            if(AdditionalData != null)
-            {
-                if(AdditionalData.TryGetValue(""@odata.nextLink"", out var nextPageLink))
-                {
-                    // Ensure it is a non empty JsonElement string
-                    if (nextPageLink is System.Text.Json.JsonElement element
-                        && element.ValueKind == System.Text.Json.JsonValueKind.String
-                        && !string.IsNullOrEmpty(element.GetString()))
-                    {
-                        Value.InitializeNextPageRequest(
-                            client,
-                            element.GetString());
-                    }
-                }
-                // Copy the additional data collection to the page itself so that information is not lost
-                Value.AdditionalData = AdditionalData;
-            }
-        }";
-}
 
 #>
--- a/Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
+++ b/Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
@@ -48,24 +48,7 @@ namespace <#=@namespace#>
             var response = await this.SendAsync<<#=collectionResponse#>>(null, cancellationToken).ConfigureAwait(false);
             if (response != null && response.Value != null && response.Value.CurrentPage != null)
             {
-                if (response.AdditionalData != null)
-                {
-                    if(response.AdditionalData.TryGetValue("@odata.nextLink", out var nextPageLink))
-                    {
-                        // Ensure it is a non empty JsonElement string
-                        if (nextPageLink is System.Text.Json.JsonElement element
-                            && element.ValueKind == System.Text.Json.JsonValueKind.String
-                            && !string.IsNullOrEmpty(element.GetString()))
-                        {
-                            response.Value.InitializeNextPageRequest(
-                                this.Client,
-                                element.GetString());
-                        }
-                    }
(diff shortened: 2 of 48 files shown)


Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
```diff
@@ -50,18 +50,18 @@ namespace <#=@namespace#>
...
-                    if (!string.IsNullOrEmpty(nextPageLinkString))
+                    if(response.AdditionalData.TryGetValue("@odata.nextLink", out var nextPageLink))
                     {
-                        response.Value.InitializeNextPageRequest(
-                            this.Client,
-                            nextPageLinkString);
+                        // Ensure it is a non empty JsonElement string
+                        if (nextPageLink is System.Text.Json.JsonElement element
+                            && element.ValueKind == System.Text.Json.JsonValueKind.String
+                            && !string.IsNullOrEmpty(element.ToString()))
```
baywet, line 58: shouldn't this be get string instead? https://docs.microsoft.com/en-us/dotnet/api/system.text.json.jsonelement.getstring?view=net-5.0#System_Text_Json_JsonElement_GetString

baywet (changes requested):

Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
andrueastman: That is valid. Especially since we have ensured that the valuekind is String.
Will action and resolve.

andrueastman: Resolved via https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/512/commits/7fa45931929c82c4696e864ad0b714110f104e2c

Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
```diff
@@ -50,18 +50,18 @@ namespace <#=@namespace#>
...
-                    var nextPageLinkString = nextPageLink as string;
-
-                    if (!string.IsNullOrEmpty(nextPageLinkString))
+                    if(response.AdditionalData.TryGetValue("@odata.nextLink", out var nextPageLink))
                     {
-                        response.Value.InitializeNextPageRequest(
-                            this.Client,
-                            nextPageLinkString);
+                        // Ensure it is a non empty JsonElement string
+                        if (nextPageLink is System.Text.Json.JsonElement element
```
baywet, line 56: is there a way we could push all the pattern matching to a single place in the "IntializeNextPageRequest" method instead of duplicating the logic?

andrueastman: I think what would happen is that instead of only giving the nextLink string to the "InitializeNextPageRequest" we would need to pass the entire response/AddtionalData for it to extract the nextLink string and create the NextPage Request object.

I think we could probably try to do it in the "CollectionResponse" object by adding a method in it that would result to something like this? (This would mean that EntityRequest.cs.tt template would remain unachanged since that template works on collection properties with customized @odata.nextlink properties)
```cs 
                var nextPageString = response.GetNextPageString(); //New method added where pattern matching happens
                if (!string.IsNullOrEmpty(nextPageString))
                {
                    response.Value.InitializeNextPageRequest(
                               this.Client,
                               nextPageString);
                }
                // Copy the additional data collection to the page itself so that information is not lost
                response.Value.AdditionalData = response.AdditionalData;
                return response.Value;
```

baywet: Why not move the `InitializeNextPageRequest` method all together into the response base class instead? this way you'll have access to everything you need to initialize the next page request and you can pass in the client.

andrueastman: @baywet Please let me know what you think of the latest update.

Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
```diff
@@ -48,24 +48,7 @@ namespace <#=@namespace#>
             var response = await this.SendAsync<<#=collectionResponse#>>(null, cancellationToken).ConfigureAwait(false);
             if (response != null && response.Value != null && response.Value.CurrentPage != null)
```
baywet, line 49: nit: this could be collapse to
```suggestion
            if (response?.Value?.CurrentPage != null)
```

Templates/templates/CSharp/Base/SharedCSharp.template.tt
```diff
@@ -334,5 +334,40 @@ public string GetMethodRequestPrimitiveReturnTypeString(string type)
...
 }
 
+/// <summary>
+/// Used in EntityCollectionResponse.cs.tt, EntityCollectionWithReferencesResponse and  MethodCollectionResponse.cs.tt
+/// to get the function to initialized the NextPageRequest object in the response of the collection request
+/// and to copy the Additional data to the CollectionPage instance
+/// </summary>
+/// <param name="iBaseClientTypeName"> type of the IBaseClient to use in creating the method</param>
+/// <returns>the string representation of the method</returns>
+public string GetInitializeCollectionPropertiesMethod(string iBaseClientTypeName)
```
baywet, line 344: could we have the collection responses derive from a base class or implement an interface with a default implementation instead of generating this all over the responses?

baywet (changes requested):
thanks for making the changes! I think we're making good progress to limit the amount of code being generated. We still have the opportunity to leverage a base class to reduce it further.

Templates/templates/CSharp/Base/SharedCSharp.template.tt
andrueastman: That makes sense. I will create a PR to the core library to expose this function for use.

andrueastman: @baywet 
When I was looking into this, I also saw that using your idea(as implemented in Java) of adding the "@odata.nextLink" as a serializable property would make the function created in the base class potentially unnecessary since the serializer would make the property available for us. Furthermore, null checking the property would also be unnecessary since the "InitializeNextPageRequest" already does this for us. We can therefore eliminate the generated method by directly passing the deserialized `NextLink` property to the "InitializeNextPageRequest".

Please check out the latest update and let me know what you think.

Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
```diff
@@ -46,26 +46,10 @@ namespace <#=@namespace#>
...
-                    var nextPageLinkString = nextPageLink as string;
-
-                    if (!string.IsNullOrEmpty(nextPageLinkString))
-                    {
-                        response.Value.InitializeNextPageRequest(
-                            this.Client,
-                            nextPageLinkString);
-                    }
-
-                    // Copy the additional data collection to the page itself so that information is not lost
```
baywet, line 65: maybe we should keep that comment for context?

Templates/templates/CSharp/Requests/EntityRequest.cs.tt
```diff
@@ -123,15 +123,17 @@ namespace <#=@namespace#>
                 {
                     <#=thisEntity_Name#>ToInitialize.<#=propertyName#>.AdditionalData = <#=thisEntity_Name#>ToInitialize.AdditionalData;
 
-                    object nextPageLink;
-                    <#=thisEntity_Name#>ToInitialize.AdditionalData.TryGetValue("<#=propertyName.ToLowerFirstChar()#>@odata.nextLink", out nextPageLink);
-                    var nextPageLinkString = nextPageLink as string;
-
-                    if (!string.IsNullOrEmpty(nextPageLinkString))
+                    if(<#=thisEntity_Name#>ToInitialize.AdditionalData.TryGetValue("<#=propertyName.ToLowerFirstChar()#>@odata.nextLink", out var nextPageLink))
```
baywet, line 126: do we still need this one or was it forgotten in the last batch of updates?

Templates/templates/CSharp/Base/SharedCSharp.template.tt
baywet: thanks for taking in the changes, I think it looks much cleaner now. A few minor things left.

Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
andrueastman: Agreed

Templates/templates/CSharp/Requests/EntityRequest.cs.tt
andrueastman: I had left this out since this was not part of the CollectionResponse but were nested collection properties of an Entity. Let me investigate the viability of doing the same for these as well and get back to you.

Templates/templates/CSharp/Requests/EntityCollectionRequest.cs.tt
andrueastman: Resolved via https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/512/commits/467f9b74e54be3ca2a8517f7af6cfe1d6039b71d

Templates/templates/CSharp/Requests/EntityRequest.cs.tt
andrueastman: This should now be resolved via https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/512/commits/6bb3ca74f2d6a844cdcec5183efec060d88b0f5c

andrueastman:
@baywet Please check out the latest updates in the PR.

baywet (approved):

PR #542, opened 2021-06-03
capture metadata with annotations

MIchaelMainer:
## Summary

Enables the capture and cleaning of metadata while maintaining capability annotations. Will add a file named cleanMetadataWithDescriptionsAndAnnotations.xml. This shouldn't change the default behavior of generator.

Results in artifacts like:
image

These generated file with capability annotations are checked in:
https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsAndAnnotationsv1.0.xml
https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/clean_beta_metadata/cleanMetadataWithDescriptionsAndAnnotationsbeta.xml

`./typewriter.exe -v Info -m https://graph.microsoft.com/v1.0/$metadata -g Transform -t https://raw.githubusercontent.com/microsoftgraph/msgraph-metadata/master/transforms/csdl/preprocess_csdl.xsl -e v1.0 -r false`

You should see instances of `Org.OData.Capabilities` in the output file.

https://github.com/microsoftgraph/msgraph-metadata/pull/70

closes https://github.com/microsoftgraph/msgraph-metadata/issues/72

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/542)


scripts/run-typewriter-clean-metadata-keep-annotations.ps1
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+
+# This script runs typewriter which results in: 1) metadata processed through an xslt and 2) documentation added as annotations in the metadata
+
+Push-Location -Path $env:TypewriterDirectory
+
+& chmod +x $env:TypewriterExecutable
```
baywet, line 7: we already have a very similar PS1. Could we update the original one instead of duplicating it so it can be used in both contexts?

src/Typewriter/Generator.cs
```diff
@@ -96,12 +96,23 @@ static internal string Transform(string csdlContents, Options options)
                 settings.EnableScript = true;
                 transform.Load(options.Transform, settings, null);
 
-                // Execute the transformation, writes the transformed file.
-                transform.Transform(doc, writer);
+                if (options.RemoveAnnotations == false)
```
baywet, line 99:
```suggestion
                if (!options.RemoveAnnotations)
```

src/Typewriter/Options.cs
```diff
@@ -81,5 +81,8 @@ public class Options
         [Option('t', "transform", HelpText = "Specify the URI to the XSLT that will preprocess the metadata. Overrides the" +
             "cleaning done by embeddeded typewriter.exe rules.")]
         public string Transform { get; set; }
+
+        [Option('r', "removeannotations", Default = (bool)true, HelpText = "Specify whether to keep source CSDL capability annotations. Only applicable when using generationMode Transform or TransformWithDocs.")]
```
baywet, line 85:
```suggestion
        [Option('r', "removeannotations", Default = true, HelpText = "Specify whether to keep source CSDL capability annotations. Only applicable when using generationMode Transform or TransformWithDocs.")]
```

test/Typewriter.Test/Given_a_valid_metadata_file_to_Typewriter.cs
```diff
@@ -560,6 +560,43 @@ public void It_transforms_metadata()
...
+            };
+
+            Generator.Transform(testMetadata, options);
+
+            FileInfo fileInfo = new FileInfo(outputDirectory + @$"{Path.DirectorySeparatorChar}{outputBaseFileName}.xml");
+            Assert.IsTrue(fileInfo.Exists, $"Expected: {fileInfo.FullName}. File was not found.");
+
+            IEnumerable<string> lines = File.ReadLines(fileInfo.FullName);
+
+            bool hasCapabilityAnnotations = false; // Expect false
```
baywet, line 586:
```suggestion
            bool hasCapabilityAnnotations = lines.Any(x => x.Contains("Org.OData.Capabilities")); // Expect false
```

test/Typewriter.Test/Given_a_valid_metadata_file_to_Typewriter.cs
```diff
@@ -560,6 +560,43 @@ public void It_transforms_metadata()
...
+
+            // Check the document for these values.
+            foreach (var line in lines)
```
baywet, line 589: remove this block

baywet (changes requested):

scripts/run-typewriter-clean-metadata-keep-annotations.ps1
MIchaelMainer: Thank you, good call. Will change.

src/Typewriter/Options.cs
MIchaelMainer: It is a requirement that we unbox when we have a default boolean value.

https://github.com/commandlineparser/commandline/wiki/Option-Attribute#default

src/Typewriter/Generator.cs
MIchaelMainer: Won't fix. I had it in this form before and I thought the that the terse form was less easy to understand.

nikithauc (approved):

PR #556, opened 2021-07-02
Fix the constructor called with no parameters add key-check for `getODataType()`

SilasKenneth:
## Summary

- Fixes the issue with missing key check before actually trying to get the key from the property dictionary for the `getODataType()` method.
- Fixes the issue with missing parameter for ctor in some classes which are actually called passing an argument.

## Generated code differences

Screenshot 2021-07-07 at 10 39 55

N/A

Issue [#287](https://github.com/microsoftgraph/msgraph-sdk-php/issues/287)
PR [#539](https://github.com/microsoftgraph/msgraph-sdk-php/pull/539)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/556)

### DIFF
--- a/Templates/templates/PHP/Model/ComplexType.php.tt
+++ b/Templates/templates/PHP/Model/ComplexType.php.tt
@@ -41,9 +41,11 @@ if (complex.IsBaseAbstractAndReferencedAsPropertyType() && !complex.IsAbstract)
 #>    /**
     * Set the @odata.type since this type is immediately descended from an abstract
     * type that is referenced as the type in an entity.
+    * @param array $propDict The property dictionary
     */
-    public function __construct()
+    public function __construct($propDict = array())
     {
+        parent::__construct($propDict);
         $this->setODataType("#<#=complex.FullName#>");
     }
 
--- a/Templates/templates/PHP/Model/EntityType.php.tt
+++ b/Templates/templates/PHP/Model/EntityType.php.tt
@@ -257,7 +257,10 @@ foreach(var property in entity.Properties.Where(prop => prop.Type.GetTypeString(
     */
     public function getODataType()
     {
-        return $this->_propDict["@odata.type"];
+        if (array_key_exists('@odata.type', $this->_propDict)) {
+            return $this->_propDict["@odata.type"];
+        }
+        return null;
     }
     
     /**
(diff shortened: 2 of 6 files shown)


Templates/templates/PHP/Model/ComplexType.php.tt
```diff
@@ -41,9 +41,11 @@ if (complex.IsBaseAbstractAndReferencedAsPropertyType() && !complex.IsAbstract)
 #>    /**
     * Set the @odata.type since this type is immediately descended from an abstract
     * type that is referenced as the type in an entity.
+    * @param array $propDict The property dictionary
     */
-    public function __construct()
+    public function __construct($propDict = array())
```
baywet, line 46: adding parameters with default values to existing methods/ctors in dotnet is compatible from a prototype standpoint but constitutes a binary breaking change. I just want to double check this is not also the case in PhP.
https://docs.microsoft.com/en-us/dotnet/core/compatibility/
https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes#binary-breaking-change

SilasKenneth: I would argue that the code wasn't working in the first place and adding this wouldn't require any code changes on the end of the client to make the new code work.

baywet: thanks for confirming

baywet (approved):

MIchaelMainer (dismissed):

Templates/templates/PHP/Model/EntityType.php.tt
```diff
@@ -257,7 +257,10 @@ foreach(var property in entity.Properties.Where(prop => prop.Type.GetTypeString(
     */
```
Ndiritu, line 257: `@return string|null`

Ndiritu (changes requested):

Ndiritu (approved):

SilasKenneth:
Hi @baywet @MIchaelMainer , kindly approve this, I made a change that dismissed your review.

baywet (approved):

PR #596, opened 2021-09-22
Fix for missing AddAsync methods for collection references

andrueastman:
Related to https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/issues/336

This PR fixes the blocked generation of POST(Add) request builders for some objects due to the expectation in the generator that posts $ref segments should have the following expectations

```cs
  /// 1. If a navigation property specifies "ContainsTarget='true'", it is self-contained. 
  ///    Generate a direct path to the item (ie "parent/child").
  /// 2. If a navigation property does not specify ContainsTarget but there is a defined EntitySet 
  ///    of the given type, it is a reference relationship. Generate a reference path to the item (ie "item/$ref").
  /// 3. If a navigation property does not have a defined EntitySet but there is a Singleton which has 
  ///    a self-contained reference to the given type, we can make a relationship to the implied EntitySet of 
  ///    the singleton. Generate a reference path to the item (ie "singleton/item/$ref").
  /// 4. If none of the above pertain to the navigation property, it should be treated as a metadata error.
```

However, the example [here](https://docs.microsoft.com/en-us/graph/api/ediscovery-sourcecollection-post-custodiansources?view=graph-rest-beta&tabs=http#request) shows that the navigation property containing the reference could be chained through several levels of contained navigation properties from the entityContainer.

This PR therefore adds a new method that can deduce if a given property is navigable from the entityContainer through a chain of navigationProperties with ContainsTarget=true. This is then used after we confirm that the four checks above are not true to avoid altering any functionality.

**Notes**
- There is the possibility of several routes (and multiple id insertions) from the entityContainer, therefore generation has been updated to let the user create the `ReferenceRequestBody` instance for themselves in the event this is exposed.

**Generations changes**
Generation Beta - https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/pull/346
Generation V1 - https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/1142

**Todo**
- [x] Validate the validity of newly exposed methods (validations added to PR descriptions in service libraries)
- [x] Update snippet generation (Resolved via https://github.com/microsoftgraph/microsoft-graph-devx-api/pull/714)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/596)

### DIFF
--- a/src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
+++ b/src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
@@ -344,7 +344,7 @@ namespace Microsoft.Graph.ODataTemplateWriter.Extensions
         private static bool IsPropertyTypeChainedContainedNavigationProperty(this OdcmProperty odcmRootProperty, OdcmProperty testProperty, HashSet<string> navigatedTypes, string route)
         {
             // check if the property already matches the type.
-            if (odcmRootProperty.Type.FullName.Equals(testProperty.Type.FullName))
+            if (odcmRootProperty.Type.FullName.Equals(testProperty.Type.FullName, StringComparison.OrdinalIgnoreCase))
             {
                 logger.Info("Property \"{0}\" matches self contained navigation property \"{1}\" of type \"{2}\"", testProperty.Name, odcmRootProperty.Name, odcmRootProperty.Type.FullName);
                 logger.Info("Possible route from service class is: {0}{1}{1}",route,Environment.NewLine);
--- a/src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
+++ b/src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
+﻿// Copyright (c) Microsoft Corporation.  All Rights Reserved.  Licensed under the MIT License.  See License in the project root for license information.
 
 namespace Microsoft.Graph.ODataTemplateWriter.Extensions
 {
@@ -352,9 +352,9 @@ namespace Microsoft.Graph.ODataTemplateWriter.Extensions
             }
 
             // check if the base class is referenced instead.
-            if (odcmRootProperty.Type.FullName.Equals(testProperty.BaseClass()?.FullName))
+            if (odcmRootProperty.AsOdcmClass()?.IsAncestorOfType(testProperty.AsOdcmClass() ?? null) ?? false)
             {
-                logger.Info("Property \"{0}\" matches self contained navigation property \"{1}\" of Base type \"{2}\"", testProperty.Name, odcmRootProperty.Name, testProperty.BaseClass()?.FullName);
+                logger.Info("Property \"{0}\" of type \"{1}\" matches self contained navigation property \"{2}\" of Base type \"{3}\"", testProperty.Name, testProperty.Type.FullName, odcmRootProperty.Name, odcmRootProperty.Type.FullName);
                 logger.Info("Possible route from service class is: {0}{1}{1}", route, Environment.NewLine);
                 return true;
             }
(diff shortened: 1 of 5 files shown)


baywet:
out of curiosity, why are PHP files getting updated by this PR? Could you double check this was intended and rebase/reset the branch if not please?

andrueastman:
Thanks for pointing this out @baywet. It was an incorrect rebase but it should be fixed now.

src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
```diff
@@ -292,6 +294,85 @@ public static OdcmProperty GetServiceCollectionNavigationPropertyForPropertyType
...
+            // check if the property already matches the type.
+            if (odcmRootProperty.Type.FullName.Equals(testProperty.Type.FullName))
+            {
+                logger.Info("Property \"{0}\" matches self contained navigation property \"{1}\" of type \"{2}\"", testProperty.Name, odcmRootProperty.Name, odcmRootProperty.Type.FullName);
+                logger.Info("Possible route from service class is: {0}{1}{1}",route,Environment.NewLine);
+                return true;
+            }
+
+            // check if the base class is referenced instead.
+            if (odcmRootProperty.Type.FullName.Equals(testProperty.BaseClass()?.FullName))
```
MIchaelMainer, line 355: Do we need to check whether any class in the base class chain is referenced instead of only the immediate base class?

andrueastman: I agree with this point. 
So far, this case is never hit though with the v1 or beta metadata. I still need to test if there are any hits if we go further up the chain. If not, I might just consider dropping this condition

andrueastman: Resolved via https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/596/commits/b1ecd0cc1114194da30039997016377ccf27c789

MIchaelMainer (approved):

andrueastman:
Thanks for taking a look at this @MIchaelMainer 

Any chance I could get an eye at the generation PRs for V1/beta as well?
- https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/pull/346
- https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/1142

src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
```diff
@@ -292,6 +294,103 @@ public static OdcmProperty GetServiceCollectionNavigationPropertyForPropertyType
...
+        /// </summary>
+        /// <param name="odcmRootProperty">The property to start the search from</param>
+        /// <param name="testProperty">The property to find in the from the root type</param>
+        /// <param name="navigatedTypes">The types already navigated to prevent circular references</param>
+        /// <param name="route">The route followed to get here</param>
+        /// <returns></returns>
+        private static bool IsPropertyTypeChainedContainedNavigationProperty(this OdcmProperty odcmRootProperty, OdcmProperty testProperty, HashSet<string> navigatedTypes, string route)
+        {
+            // check if the property already matches the type.
+            if (odcmRootProperty.Type.FullName.Equals(testProperty.Type.FullName))
```
baywet, line 347:
```suggestion
            if (odcmRootProperty.Type.FullName.Equals(testProperty.Type.FullName, StringComparison.OrdinalIgnoreCase))
```

baywet (changes requested):

src/GraphODataTemplateWriter/Extensions/OdcmModelExtensions.cs
andrueastman: Thanks, this is now resolved via https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/596/commits/a0a850149876d7f613aede4c228677f706a63978

PR #645, opened 2021-11-18
Disambiguation for Collection Entities in Java

ramsessanchez:
Fixing conflicting class naming issue causing latest Java Beta Build to fail. 
Currently there is no disambiguation for Entities ending in Collection. This has caused an issue because an Entity titled 'someName' will have several conflicting generated classes with the Entity 'someNameCollection'. This is especially true when collections of both entities exist. A current example is the following:
ManagementTenant creates the class ManagementTemplateCollectionRequest via the template '<BaseEntity>CollectionRequest'
ManagementTenantCollection also wants to create the class ManagementTemplateCollectionRequest, however, via the template '<BaseEntity>Request'. 
This issue applies to other similarly named Templates. 
By adding the term 'Object' to any Entity with the suffix 'Collection' we will be able to differentiate the two and avoid future naming conflicts.  
Now, in our example, we will have ManagementTenantCollectionRequest & ManagementTenantCollectionObjectRequest. 
https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/204

https://github.com/microsoftgraph/msgraph-sdk-java/pull/942 -- V1 Potential Changes
https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/207 -- Beta Potential Changes


MIchaelMainer:
Can we make this more targeted so that it only affects ManagementTenant and ManagementTemplateCollection entities? The change as is would impact v1.0 with breaking changes. Or were those scenarios with changes in v1.0 already broken and this fixes it?

baywet (changes requested):
I agree with Michael on the fact this change has too broad of an impact. [more details here](https://github.com/microsoftgraph/msgraph-sdk-java/pull/942#pullrequestreview-811237556)
Besides scoping the change a bit more (understanding why the sitecollection entity type is not causing an issue for instance), @darrelmiller is looking to see if he can have the workload change the name they are using since their naming is really confusing anyway.
Let's keep iterating on a solution here to narrow it down to a smaller impact in the meantime.

ramsessanchez:
Edited the naming condition to target ManagementTenantCollection specifically. Currently builds successfully locally. Will trigger a beta build using latest metadata.

ramsessanchez:
https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/219

baywet (approved):

baywet:
Also, this is probably the first contribution you're making to this repo. For those changes to be accounted for in the weekly pipeline you need to open a PR from dev to master.

PR #740, opened 2022-05-09
[Dotnet] Remove abstract modifier for complex types

andrueastman:
This PR is part of https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/1328

It updates the Typewriter generator to behave consistently for both complex and abstract types by removing the abstract modifier on complex types.

This therefore unblocks instances where new derived types of abstract complex types are added to the metadata leading to the throwing of an exception. 

With this change, the derivedTypeConverter should be able to instantiate and return an instance of the base complex type in the event the user has not updated to a version where the derived type is not present in the metadata.

Generation PRs
V1 - https://github.com/microsoftgraph/msgraph-sdk-dotnet/pull/1340
Beta - https://github.com/microsoftgraph/msgraph-beta-sdk-dotnet/pull/443

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/740)

### DIFF
--- a/Templates/templates/CSharp/Model/ComplexType.cs.tt
+++ b/Templates/templates/CSharp/Model/ComplexType.cs.tt
@@ -28,7 +28,6 @@ if (complex.Base != null)
 }
 
 var isMethodResponse = complex.LongDescriptionContains("methodResponse");
-var classType = complex.IsAbstract ? "abstract partial class" : "partial class";
 
 var attributeStringBuilder = new StringBuilder();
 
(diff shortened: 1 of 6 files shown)


baywet (approved):

andrueastman:
Hey @zengin,

The PR makes the behavior of abstract entity and abstract complex types consistent due to failures observed when the API returns new complex types derived from abstract classes. So, I believe we should still be able to capture the errors as the constructor is marked as `protected internal` like the entity types meaning that it can only be called in the same assembly (by the serializer during deserialization to prevent crashing) or by a derived type.

For Kiota, there is no distinction at the moment for abstract and non-abstract types.

zengin:
@andrueastman that makes sense. Thanks for highlighting the constructor signature.

PR #794, opened 2022-08-04
Rsh/java fully qualified name

ramsessanchez:
This PR covers the case of adding a fully qualified name to entity properties and entity-collection properties in instances where two entities of the same name exist in different namespaces but are referenced in the same class. To prevent collisions and disambiguate the properties which share a name, they are now referenced by their fully qualified name. 
Furthermore, entities with the same name cannot both be present in import statements as they have the same 'simple name' which is not allowed in java. For this reason, in cases where two entities with the same name appear, we have to remove the import statements as well as reference them by their fully qualified name. This is allowed in java since no import statement is required when a class is referenced by its fully qualified name. 
Lastly, this pr also contains work to remove many unused import statements. Currently we generate import statements for the models which are referenced by a collection, i.e., a if there is a collection of 'EntityZ' there will be import statements for the entity collection, its requests, and its builders, however there will also be an import for EntityZ even though it is not referenced directly and thus does not need to be imported. This was an issue in the Security entity class as it had collections of two different Alert entities which resulted in several import collision conflicts, as well as many unused imports. 
https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/352

extra: many methods which were not being used at all have been removed from the java type helper. 
 
 note: This does not cover every case in which collisions may appear, nor does it add a fully qualified name to every property, model, collection of models, or etc. However, this does cover the current scenarios preventing https://github.com/microsoftgraph/msgraph-beta-sdk-java/tree/beta/pipelinebuild/81419 from building. As more cases of deprecation occur, as we see now with Alerts and Alerts_V2, there will be much more work needed to be done with how we reference entities which share a name. There are many cases to check and possibly fix, however in the interest of time and with the release of Kiota soon, I am choosing to not address these unless we have to.


src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
```diff
@@ -944,10 +890,28 @@ public static string CreatePackageDefForEntity(this CustomT4Host host)
...
+            var collidedProperties = new HashSet<string>();
+            var allProperties = new HashSet<string>();
+            foreach(var property in properties)
+            {
+                var rawTypeName = property.Projection.Type.Name;
+                if (allProperties.Contains(rawTypeName))
+                {
+                    collidedProperties.Add(rawTypeName);
+                }
+                allProperties.Add(rawTypeName);
```
baywet, line 902: else block

baywet (dismissed):
Thanks for taking this on, please also revert the vipr submodule change, I believe your submodule was behind when you committed. (git submodule update --remote --merge)
For some reason I see a lot of whitespace changes in the PR you put together here? https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/351

src/GraphODataTemplateWriter/CodeHelpers/Java/TypeHelperJava.cs
ramsessanchez: > Thanks for taking this on, please also revert the vipr submodule change, I believe your submodule was behind when you committed. (git submodule update --remote --merge) For some reason I see a lot of whitespace changes in the PR you put together here? [microsoftgraph/msgraph-beta-sdk-java#351](https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/351)
Also seeing the whitespace, I believe this may be an issue on the github side. If we look at the line numbers, they still match up for files which are unchanged and off by a couple for those which have a few line changes. Not sure why the empty space is being shown.

ramsessanchez:
> Thanks for taking this on, please also revert the vipr submodule change, I believe your submodule was behind when you committed. (git submodule update --remote --merge) For some reason I see a lot of whitespace changes in the PR you put together here? [microsoftgraph/msgraph-beta-sdk-java#351](https://github.com/microsoftgraph/msgraph-beta-sdk-java/pull/351)

Also seeing the whitespace, I believe this may be an issue on the github side. If we look at the line numbers, they still match up for files which are unchanged and off by a couple for those which have a few line changes. Not sure why the empty space is being shown.

zengin (changes requested):
@ramsessanchez let's avoid having the whitespace change in the SDK code history. I'd be happy to help if you'd like to take a look at it together.

zengin:
@ramsessanchez I had a quick look at the diff locally, it looks like the issue is having `\r\n` instead of `\n` in some cases, but on further investigation, it looks like Typewriter templates have it mixed at the moment. So I will not block this PR on that.

zengin (dismissed):

zengin (approved):

PR #835, opened 2022-09-22
Adds a base client for go generation

rkodev:
## Summary

Renames default client generated by go from `GraphServiceClient` to `GraphBaseServiceClient`. This should allow adding manually edited files later.

Partially addresses https://github.com/microsoftgraph/msgraph-sdk-go-core/issues/88

###
Blocked by 
https://github.com/microsoftgraph/msgraph-sdk-go/pull/277
https://github.com/microsoftgraph/msgraph-beta-sdk-go/pull/155

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/835)

### DIFF
--- a/scripts/clean-go-files.ps1
+++ b/scripts/clean-go-files.ps1
@@ -1,7 +1,7 @@
-$directories = Get-ChildItem -Path $env:MainDirectory -Directory -Exclude @("core", "scripts", ".github")
+$directories = Get-ChildItem -Path $env:MainDirectory -Directory -Exclude @("core", "scripts", ".github" , "tests")
 foreach ($directory in $directories) {
 	Remove-Item -Path $directory.FullName -Recurse -Force -Verbose
 }
-Get-ChildItem -Path $env:MainDirectory -Filter graph_base_service_client.go -Recurse | Remove-Item
+Remove-Item -Path $env:MainDirectory -Filter "*.go" -Exclude "graph_request_adapter.go", "graph_service_client.go" -Verbose
 
 Write-Host "Removed the existing generated files in the repo's main directory: $env:MainDirectory" -ForegroundColor Green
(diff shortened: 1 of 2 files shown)


scripts/clean-go-files.ps1
```diff
@@ -2,6 +2,6 @@ $directories = Get-ChildItem -Path $env:MainDirectory -Directory -Exclude @("cor
 foreach ($directory in $directories) {
 	Remove-Item -Path $directory.FullName -Recurse -Force -Verbose
 }
-Remove-Item -Path $env:MainDirectory -Filter "*.go" -Exclude "graph_request_adapter.go" -Verbose
+Get-ChildItem -Path $env:MainDirectory -Filter graph_base_service_client.go -Recurse | Remove-Item
```
baywet, line 5: Remove-Item -Path $env:MainDirectory -Filter "*.go" -Exclude "graph_request_adapter.go", "graph_service_client.go" -Verbose

baywet (changes requested):

scripts/clean-go-files.ps1
rkodev: My only concern is that we will be introducing [tests](https://github.com/microsoftgraph/msgraph-sdk-go/blob/feat/simplify-client-initialization/graph_service_client_test.go) , as well as possibly some utility methods. i.e helper methods for the batch requests. Given that the generation only generates 1 file while all the other top level files  are manually written, wouldn't it be easier specify the file to delete rather than files to keep

baywet: ok but the current script won't remove anything else but the generated client, not the models or the request builders. Which we want to remove for sanity. Maybe have an array defined above for the exclude values entries? We could also move the tests to a subdirectory I believe if that makes things easier.

rkodev: moved the tests to a separate folder

baywet (approved):
Thanks for making the changes

PR #925, opened 2023-02-03
PHP Entity bug : Test never evaluates to true

AntoineMarques:
Targets PHP generated Entity model class.

Resolution of class `\Entity` is the problem.
Running `class_exists("\Entity")` returns `false` while `class_exists("\MicrosoftGraph\Core\Model\Entity")` is true. 
Because of beta api, the most simple fix is using `self::class`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/925)

### DIFF
--- a/Templates/templates/PHP/Model/EntityType.php.tt
+++ b/Templates/templates/PHP/Model/EntityType.php.tt
@@ -291,7 +291,7 @@ foreach(var property in entity.Properties.Where(prop => prop.Type.GetTypeString(
                 $serializableProperties[$property] = $val->format(\DateTime::RFC3339);
             } else if (is_a($val, "\Microsoft\Graph\Core\Enum")) {
                 $serializableProperties[$property] = $val->value();
-            } else if (is_a($val, "\Entity")) {
+            } else if (is_a($val, self::class)) {
                 $serializableProperties[$property] = $val->jsonSerialize();
             } else if (is_a($val, "\GuzzleHttp\Psr7\Stream")) {
                 $serializableProperties[$property] = (string) $val;


Ndiritu:
Thanks for catching this bug @AntoineMarques.
Seems like serialization has been working since `json_encode()` recursively serializes the properties anyway. Might there be a bug we're unaware of?

What are your thoughts on removing the check for `Entity` altogether?

cc: @SilasKenneth

SilasKenneth:
> Thanks for catching this bug @AntoineMarques.
> 
> Seems like serialization has been working since `json_encode()` recursively serializes the properties anyway. Might there be a bug we're unaware of?
> 
> @SilasKenneth Thinking we can remove the checks for properties that are `Entity` types altogether. Thoughts?

What could possibly go wrong with removing the entity check? I suspect we might run into a few issues. I believe we can try a fix(or remove it) and see the effect. From the look of the code, I suspect it might always evaluate to false, especially if  the class it is called from is not in the same namespace as Entity. Maybe we can remove the quotes and say `Entity::class` instead of `"\Entity"` or even `self::class`.

Ndiritu:
Yes, `self::class` should be the accurate check as in the PR.

However, I think the accuracy & efficiency of serializing nested entities is better left to `json_encode`? (As is the case now)

Making our check pass introduces a higher risk of bugs since we will change the order of serialization of nested properties & `json_encode` will still recurse through the entire object structure anyway?

baywet:
Closing since the [newer version](https://github.com/microsoftgraph/msgraph-sdk-php/releases/tag/2.0.0) of the SDK now relies on [kiota](https://aka.ms/kiota) for generation.

PR #944, opened 2023-02-22
Feature: generate multiple OpenAPI definitions

thewahome:
## Summary

The DevX API, when it receives a request, performs the conversion from CSDL to OpenAPI before returning the converted open api.  This is done in-process. Optimisations have been made to use caching however the conversion process takes time.

This work intends to do the conversion at source so that the DevX API's work is reduced to pure fetching of the document and fulfilling the requests made.

> Conversion settings

I've introduced a `conversion-settings` folder. It holds the OpenAPI conversion settings that will be used to override the default Hidi settings when performing the `publish openapi` build step. 
The pipeline reads through the folder and gets the filenames and spawns new jobs to perform the publish based on the azure pipelines matrix strategy. 
This is going to be useful when it comes to maintenance since the only changes to the generation process happen in this folder. Any new settings are immediately picked up. Deleted ones just reduce the eventually spawned jobs.

## Generated code differences

Once published the files are pushed to the `microsoftgraph/msgraph-metadata` repository and creates a folder structure that looks like this:

    openapi/
    ├── v1.0/
    └── beta/                    # Scoped to beta for brevity. The v1.0 folder should share the same folder structure
        ├── openapi.yml    
        ├── graphexplorer.yml         
        ├── powershell.yml 
        └── ...                # any other introduced platform
    

No new command lines introduced

### DIFF
--- a/.azure-pipelines/generation-pipeline.yml
+++ b/.azure-pipelines/generation-pipeline.yml
@@ -154,6 +154,7 @@ variables:
   kiotaDirectory: '$(Build.SourcesDirectory)/kiota'
   typewriterExecutable: '$(typewriterDirectory)/Typewriter'
   scriptsDirectory: '$(Build.SourcesDirectory)/MSGraph-SDK-Code-Generator/scripts'
+  conversionSettingsDirectory: '$(Build.SourcesDirectory)/MSGraph-SDK-Code-Generator/conversion-settings'
   transformScript: '$(Build.SourcesDirectory)/msgraph-metadata/transforms/csdl/preprocess_csdl.xsl'
   docsDirectory: '$(Build.SourcesDirectory)/microsoft-graph-docs'
 
@@ -224,25 +225,21 @@ stages:
   dependsOn: stage_v1_metadata
   condition: in(dependencies.stage_v1_metadata.result, 'Succeeded', 'Skipped')
   jobs:
-  - job: v1_openapi
-    steps:
-    - template: generation-templates/capture-openapi.yml
-      parameters:
-        endpoint: 'v1.0'
-        outputPath: $(cleanOpenAPIFileV1OutputPath)
-        cleanMetadataFolder: $(cleanOpenAPIFolderV1)
+  - template: generation-templates/capture-openapi.yml
+    parameters:
+      endpoint: 'v1.0'
+      outputPath: $(cleanOpenAPIFileV1OutputPath)
+      cleanMetadataFolder: $(cleanOpenAPIFolderV1)
 
 - stage: stage_beta_openapi
   dependsOn: stage_beta_metadata
   condition: in(dependencies.stage_beta_metadata.result, 'Succeeded', 'Skipped')
   jobs:
-  - job: beta_openapi
-    steps:
-    - template: generation-templates/capture-openapi.yml
-      parameters:
-        endpoint: 'beta'
-        outputPath: $(cleanOpenAPIFileBetaOutputPath)
-        cleanMetadataFolder: $(cleanOpenAPIFolderBeta)
+  - template: generation-templates/capture-openapi.yml
+    parameters:
+      endpoint: 'v1.0'
+      outputPath: $(cleanOpenAPIFileV1OutputPath)
+      cleanMetadataFolder: $(cleanOpenAPIFolderV1)
 
 - stage: stage_csharp_v1
   dependsOn:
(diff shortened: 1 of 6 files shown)


.azure-pipelines/generation-templates/capture-openapi.yml
```diff
@@ -13,55 +13,86 @@ parameters:
...
+      git fetch origin master
+      git switch master
+    displayName: ensure the generation happens from master latest
+    workingDirectory: $(Build.SourcesDirectory)/msgraph-metadata
 
-- pwsh: |
-    ./scripts/run-openapi-validation.ps1 -repoDirectory (Get-Location).Path -version "${{ parameters.endpoint }}"
-  displayName: ensure that OpenAPI docs can be parsed
-  workingDirectory: $(Build.SourcesDirectory)/msgraph-metadata
+  - pwsh: '$(scriptsDirectory)/generate-open-api.ps1 -endpointVersion ${{ parameters.endpoint }} -settings "$(conversionSettingsDirectory)/$(File)" -platformName "$(Name)"'
```
baywet, line 68: where is the File variable supposed to come from?

baywet (changes requested):

.azure-pipelines/generation-templates/capture-openapi.yml
thewahome: When the pipeline iterates through the conversion-settings folder, it places the values in a variable. Then each spawned job has access to the File variable.

baywet: are you referring to this? https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/944/files#diff-aebcf22971e4a4bda2ad75e658cf157112fa98a04cb56c4d186e8e03b94e8998R29

baywet: thanks for clarifying!

.azure-pipelines/generation-pipeline.yml
```diff
@@ -224,25 +225,21 @@ stages:
...
-  - job: beta_openapi
-    steps:
-    - template: generation-templates/capture-openapi.yml
-      parameters:
-        endpoint: 'beta'
-        outputPath: $(cleanOpenAPIFileBetaOutputPath)
-        cleanMetadataFolder: $(cleanOpenAPIFolderBeta)
+  - template: generation-templates/capture-openapi.yml
+    parameters:
+      endpoint: 'v1.0'
```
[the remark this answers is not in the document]
thewahome: If I commit this suggestion we will have 2 beta generations

baywet:
Thanks for making the changes, the only thing left would be to move the settings from this repo to the metadata one and make related updates here, but after that we should be good to go.

thewahome:
> Thanks for making the changes, the only thing left would be to move the settings from this repo to the metadata one and make related updates here, but after that we should be good to go.

You probably hit sent while this last change was syncing

.azure-pipelines/generation-pipeline.yml
```diff
@@ -154,6 +154,7 @@ variables:
   kiotaDirectory: '$(Build.SourcesDirectory)/kiota'
   typewriterExecutable: '$(typewriterDirectory)/Typewriter'
   scriptsDirectory: '$(Build.SourcesDirectory)/MSGraph-SDK-Code-Generator/scripts'
+  conversionSettingsDirectory: '$(Build.SourcesDirectory)/MSGraph-SDK-Code-Generator/conversion-settings'
```
baywet, line 157: this needs to be updated as well, And the powershell script could rely on this variable instead of building the path itself

baywet (changes requested):

.azure-pipelines/generation-pipeline.yml
thewahome: This surprisingly failed in my tests. Having it like this was the only way it passed

baywet (approved):
Thanks for making the changes

.azure-pipelines/generation-pipeline.yml
baywet: right because variables don't survive between jobs, thanks for fixing the original one

.azure-pipelines/generation-templates/capture-openapi.yml
```diff
@@ -13,55 +13,87 @@ parameters:
...
+      foreach ($f in Get-ChildItem $dirPath -Recurse -Include *.json) {
+        $name = $f.Name.replace('.json','')
+        $data = @{"File"=$f.Name;"Name"=$name;}
+        $key = "$endpoint-$name"
+        $jsonBase.Add($key,$data)
+      }
+      $json = $jsonBase | ConvertTo-Json -Compress
+      Write-Host "##vso[task.setvariable variable=targets;isOutput=true]$json"
+    name: setTargets
+    workingDirectory: $(Build.SourcesDirectory)/msgraph-metadata
```
andrueastman, line 36:
```yml
- template: checkout-metadata.yml
```
Working directory does not exist yet. Line 56 will need to be called before this.

PR #1085, opened 2023-10-31
Update base directory name for beta sdk

samwelkanda:
Specifies the base directory name for the python v1.0 and beta sdks.


.azure-pipelines/generation-templates/python-beta.yml
```diff
@@ -0,0 +1,17 @@
+parameters:
+  - name: repoName
+    type: string
+
+steps:
+- pwsh: '$(scriptsDirectory)/clean-python-files.ps1'
+  displayName: 'Remove generated models and requests from the repo generated folder'
+  env:
+    RepoModelsDir: '$(Build.SourcesDirectory)/${{ parameters.repoName }}/msgraph_beta/generated'
```
baywet, line 9: my understanding is that we duplicated the file just for the msgraph_beta segment here (x2), could we not instead add an additional parameter to original template?

baywet (changes requested):

andrueastman (approved):

baywet (approved):
Thanks for making the changes

PR #1113, opened 2023-11-24
Updates beta metadata source

irvinesunday:
Fixes https://github.com/microsoftgraph/msgraph-metadata/issues/203

Switches the source of beta metadata from the live site (https://graph.microsoft.com/beta/$metadata) to the CSDL files in the _./schema_ folder in the metadata repo. This file gets maintained by the Identity folks and it is an annotation-enriched version of the live site file.

We are validating and testing this move with the `beta` metadata first then switch over v1.0 later.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/1113)

### DIFF
--- a/.azure-pipelines/generation-templates/capture-metadata.yml
+++ b/.azure-pipelines/generation-templates/capture-metadata.yml
@@ -47,6 +47,7 @@ steps:
     errorOnNoChange: 1
     endpointVersion: ${{ parameters.endpoint }}
     targetBranch: 'master'
+    InputMetadataFile: ${{ parameters.inputMetadata }}
   enabled: true
 
 - template: download-typewriter.yml
--- a/.azure-pipelines/generation-templates/capture-metadata.yml
+++ b/.azure-pipelines/generation-templates/capture-metadata.yml
@@ -47,7 +47,7 @@ steps:
     errorOnNoChange: 1
     endpointVersion: ${{ parameters.endpoint }}
     targetBranch: 'master'
-    InputMetadataFile: ${{ parameters.inputMetadata }}
+    inputMetadataFile: ${{ parameters.inputMetadata }}
   enabled: true
 
 - template: download-typewriter.yml
--- a/scripts/download-diff-metadata.ps1
+++ b/scripts/download-diff-metadata.ps1
@@ -49,7 +49,7 @@ if ($branch -ne $env:targetBranch) {
     git pull origin $env:targetBranch --ff-only | Write-Host
 }
 
-if ($env:inputMetadataFile.StartsWith("http")){    
+if ($env:inputMetadataFile.StartsWith("http", [StringComparison]::OrdinalIgnoreCase)){    
     # Download metadata from livesite.
     $url = "https://graph.microsoft.com/{0}/`$metadata" -f $env:endpointVersion
     $metadataFileName = "{0}_metadata.xml" -f $env:endpointVersion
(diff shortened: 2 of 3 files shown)


scripts/download-diff-metadata.ps1
```diff
@@ -49,20 +49,28 @@ if ($branch -ne $env:targetBranch) {
...
-# Download the metadata from livesite.
-$url = "https://graph.microsoft.com/{0}/`$metadata" -f $env:endpointVersion
-$metadataFileName = "{0}_metadata.xml" -f $env:endpointVersion
-$pathToLiveMetadata = Join-Path -Path ($pwd).path -ChildPath $metadataFileName
-$client = new-object System.Net.WebClient
-$client.Encoding = [System.Text.Encoding]::UTF8
-Write-Host "Attempting to downloaded metadata from $url to $pathToLiveMetadata" -ForegroundColor DarkGreen
-$client.DownloadFile($url, $pathToLiveMetadata)
-Write-Host "Downloaded metadata from $url to $pathToLiveMetadata" -ForegroundColor DarkGreen
+if ($env:endpointVersion -eq "beta"){
```
andrueastman, line 52: Rather than evaluating the endpoint version, maybe we should consider having the script simply evaluate if `$env:InputMetadataFile` is a URI or a file path and then use the result to set the `$content`?
That way when we turn this generation source on for v1, we will only need to change the environment variable rather than changing/modifying this script again?

baywet (commented):
Why only swap beta and not v1 as well?

sebastienlevert:
@baywet 

> We are validating and testing this move with the beta metadata first then switch over v1.0 later.

.azure-pipelines/generation-templates/capture-metadata.yml
```diff
@@ -47,6 +47,7 @@ steps:
     errorOnNoChange: 1
     endpointVersion: ${{ parameters.endpoint }}
     targetBranch: 'master'
+    InputMetadataFile: ${{ parameters.inputMetadata }}
```
baywet, line 50: nit, the casing differs from other parameters

baywet (commented):
Thanks @sebastienlevert I wasn't fully up yet apparently...

scripts/download-diff-metadata.ps1
```diff
@@ -49,14 +49,8 @@ if ($branch -ne $env:targetBranch) {
...
 
-if ($env:endpointVersion -eq "beta"){
-    # Retrieve the beta metadata from ./schemas folder in metadata repo
-    Write-Host "Retrieving metadata from $env:InputMetadataFile" -ForegroundColor DarkGreen
-    $content = Format-Xml (Get-Content $env:InputMetadataFile)
-    Write-Host "Retrieved metadata from $env:InputMetadataFile" -ForegroundColor DarkGreen
-}
-else{ 
-    # Download the v1.0 metadata from livesite.
+if ($env:inputMetadataFile.StartsWith("http")){
```
baywet, line 52:
```suggestion
if ($env:inputMetadataFile.StartsWith("http", [StringComparison]::OrdinalIgnoreCase)){    
```

baywet (approved):

scripts/download-diff-metadata.ps1
irvinesunday: Thanks

PR #1141, opened 2024-01-19
Revert "Switches `v1.0` metadata source from livesite to local CSDL file in metadata repo."

baywet:
Reverts microsoftgraph/MSGraph-SDK-Code-Generator#1140

(temp revert of that change because it seems to be causing over-expansion, currently investigating)
https://dev.azure.com/microsoftgraph/Graph%20Developer%20Experiences/_build/results?buildId=131797&view=results
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/1141)

### DIFF
--- a/.azure-pipelines/generation-pipeline.yml
+++ b/.azure-pipelines/generation-pipeline.yml
@@ -159,7 +159,7 @@ variables:
   cleanOpenAPIFolderBeta: 'clean_beta_openapi'
   cleanOpenAPIFolderV1: 'clean_v10_openapi'
   rawMetadataFileBeta: '$(Build.SourcesDirectory)/msgraph-metadata/schemas/beta-Prod.csdl'
-  rawMetadataFileV1: '$(Build.SourcesDirectory)/msgraph-metadata/schemas/v1.0-Prod.csdl'
+  rawMetadataFileV1: 'https://graph.microsoft.com/v1.0/$metadata' # We want to run against the metadata we have captured.
   typewriterDirectory: '$(Build.SourcesDirectory)/typewriter'
   kiotaDirectory: '$(Build.SourcesDirectory)/kiota'
   typewriterExecutable: '$(typewriterDirectory)/Typewriter'


baywet:
@irvinesunday after some more investigation I believe the problem is actually in hidi and more specifically in the conversion library.
Running the pipeline with this change reverted didn't make it succeed.
Can you make it a priority to investigate this over-expansion issue on Monday please?

irvinesunday:
> @irvinesunday after some more investigation I believe the problem is actually in hidi and more specifically in the conversion library. Running the pipeline with this change reverted didn't make it succeed. Can you make it a priority to investigate this over-expansion issue on Monday please?

Noted

irvinesunday:
Seems the issue is the [composable functions](https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/1141#issuecomment-1900994382) feature that was added in the conversion lib. sometime back in early Dec. 2023, specifically bloating up the `drives` API.

Here's the diff in paths: https://www.diffchecker.com/5Xb89XRx/

We might need to revert this feature in the interim, so as to unblock the weekly generation, while we investigate further.

andrueastman:
> We might need to revert this feature in the interim, so as to unblock the weekly generation, while we investigate further.

Taking a look at https://github.com/microsoft/OpenAPI.NET.OData/pull/431/files, we may have to revert it in the short term. Would it be make sense to bring it back with a setting/configuration so that it can still be taken advantage of?

irvinesunday:
Having set the composability of functions with all the return type of `graph.workbookRange` to `false` in the metadata, the number of paths and OpenAPI size reduce significantly. Here's a diff. of the the number of paths generated with that change: https://www.diffchecker.com/hjHW2WRx/

image

The effect of setting the composabilities of functions whose return type is `graph.workbookRange` to `false` ensures we don't generate the navigation properties for this type, which is the cause of the bloated OpenAPI paths and size. 
The OpenAPI file reduces to 40MB (for v1.0) as a consequence of this. 
In order to sort this, we will need to:

 1) We need to go ahead and make this change in our XSL transform, by setting `IsComposable="false"` to all functions whose  `ReturnType Type="graph.workbookRange"`

 2) [_This is not priority however_] We may need to make an update to the generation of `OperationId`s of navigation properties of composable functions in the conversion lib. as we did not account for this and we have duplicate operation ids been generated. However, since PowerShell is the only client that relies on OperationIds (courtesy of AutoREST) and these composable paths under the `drives` API don't show up in the PowerShell/GE OpenAPI files, this is not a required fix to unblock the generation, and can be prioritized later.

andrueastman:
I'm happy with this choice. Missing paths/function are exposed while functions `graph.workbookRange` should still be usable in the SDKs similar to how we handle driveItems in https://github.com/microsoftgraph/msgraph-metadata/issues/248

irvinesunday:
This PR should fix this: https://github.com/microsoftgraph/msgraph-metadata/pull/561

baywet:
Thanks everyone for coming together. Closing this since the source of the issue is not switching to the livesite metadata.

PR #1350, opened 2025-04-08
fix: metadata validation is missing hidi installation

baywet:
Signed-off-by: Vincent Biret <vibiret@microsoft.com>

### DIFF
--- a/.azure-pipelines/generation-templates/capture-metadata.yml
+++ b/.azure-pipelines/generation-templates/capture-metadata.yml
@@ -52,7 +52,10 @@ steps:
 
 - template: download-typewriter.yml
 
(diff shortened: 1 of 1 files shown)


.azure-pipelines/generation-templates/capture-metadata.yml
```diff
@@ -115,6 +115,14 @@ steps:
 
 - template: use-dotnet-sdk.yml
 
+# required for the hidi installation validation
+- template: use-dotnet-sdk.yml
```
gavinbarron, line 119: don't you need to remove the template entry on line 116?

gavinbarron (approved):

PR #1442, opened 2026-06-25
ci: use GitHub App identity instead of PAT for Agents-M365Copilot

gavinbarron:
## Summary

Pivots the **Agents-M365Copilot** generation flow off the PAT-backed `microsoftkiota` GitHub service connection and onto the existing `microsoft-graph-devx-bot` **GitHub App** identity. PR creation already used the App, so this completes PAT elimination for Agents-M365Copilot checkout/push/PR.

Because `microsoft/Agents-M365Copilot` is a public repo, the checkout is done anonymously (no credential) and the App token is used **only** for the push.

## Changes
- **`scripts/set-app-token-push-url.ps1`** (new): generates a GitHub App installation token (reusing `Generate-Github-Token.ps1`), masks it, and sets only the `origin` *push* URL to use it. Fetch stays anonymous.
- **`language-generation-kiota.yml`**: added opt-in `useGitHubAppAuth` param (default `false`). When `true` → anonymous clone replaces the resource checkout; AKV secrets + push-URL step run before the push; existing `git-push-files.ps1` / `create-pull-request.ps1` unchanged. When `false` → microsoftgraph SDK repos behave exactly as before.
- **`generation-pipeline.yml`**: removed the `Agents-M365Copilot` repo resource (the `microsoftkiota` endpoint) and its SDL-exclude entry; set `useGitHubAppAuth: true` on all 6 Agents jobs (csharp/python/typescript × v1/beta).

## Validation
- No `microsoftkiota` references remain; new script parses clean; no dangling checkout references; SDK-repo path untouched (param defaults false).

## Note
- The `microsoft-graph-devx-bot` App must have `contents:write` on `microsoft/Agents-M365Copilot` (PR creation already proves install + PR perms).

> Automated change authored with GitHub Copilot CLI.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/1442)

PR #1447, opened 2026-07-10
docs(ci): clarify why GitHub App auth is used for cross-org repos

ramsessanchez:
## Summary

Follow-up to #1442. Clarifies the comments/documentation around the `useGitHubAppAuth` path so the distinctions we landed on are explicit and not misleading.

> Note: this is a doc-only follow-up to #1442. The original comments implied the target repo being **public** is what drives the switch to GitHub App auth. It isn't - the other SDK repos are public too. The real driver is that the target repo lives in a **different GitHub org** than the shared service connection covers, which historically required its own PAT-backed connection; the flag removes that PAT by reusing the GitHub App identity that already opens the PRs. Public visibility only enables the anonymous, credential-free clone - it is not why App auth is used.

## Changes
- **`language-generation-kiota.yml`**: rewrote the `useGitHubAppAuth` param comment to lead with the cross-org / PAT-elimination rationale and add an explicit note that repo visibility is not the selector; reworded the anonymous-clone step comment to frame public visibility as the enabler of credential-free reads rather than the reason for App auth.
- **`scripts/set-app-token-push-url.ps1`**: reframed the header so push-only URL rewriting reads as a deliberate optimization for public repos, not a requirement of App auth.

No behavioral changes - comments/documentation only.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoftgraph/MSGraph-SDK-Code-Generator/pull/1447)

### DIFF
--- a/.azure-pipelines/generation-templates/language-generation-kiota.yml
+++ b/.azure-pipelines/generation-templates/language-generation-kiota.yml
@@ -61,12 +61,6 @@ parameters:
 # needed their own separate PAT-backed one. This flag removes that PAT by reusing the
 # GitHub App identity (which already creates the PRs) for the push as well.
 #
-# NOTE: repo visibility is NOT what selects this path - the SDK repos on the default path
-# are public too. Visibility only affects HOW the target is read: when the repo is public
-# we can clone it anonymously and avoid needing any read credential at all, leaving the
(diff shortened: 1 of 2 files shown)


ramsessanchez:
comments only change, merging

